Issue #9, opened 2023-08-09
Create docs: WIT reference documentation


kate-goldenring:
Per [proposed TOC](https://github.com/bytecodealliance/component-docs/pull/1), add section that includes:

More readable version of [the WIT Definition Language doc](https://github.com/WebAssembly/component-model/blob/main/design/mvp/WIT.md)

lann:
If its useful, I wrote up a WIT quick ref: https://github.com/lann/wit-quick-ref

itowlson:
I can't assign myself _whee_ but I am working on this.

PR #14, opened 2023-08-14
WIT overview

itowlson:
WIP at the moment.

It's getting quite long; I'm wondering if we should split it into multiple pages?  Or perhaps one page that summarises the key concepts and some common syntax, and then the rest of the content can move to a "reference but the friendly kind" section?


component-model/src/wit-overview.md
```diff
@@ -0,0 +1,342 @@
...
+option<customer>
+```
+
+This is similar to Rust `Option`, C++ `std::optional`, or Haskell `Maybe`.
+
+> This is a special case of a [variant](#variants) type.  WIT defines it so that there is a common way of expressing it, so that you don't need to create a record type for every value type, and to enable it to be mapped idiomatically into languages with option types.
+
+### Results
+
+`result<T, E>` for any types T and E may contain a value of type T _or_ a value of type E (but not both). This is typically used for "value or error" situations; for example, a HTTP request function might return a result, with the success case (the T type) representing a HTTP response, and the error case (the E type) representing the various kinds of error that might occur:
```
lann, line 101: Notably, you can also omit either type, e.g. `result<T>`, `result<_, E>` or just `result` (omitting both).

component-model/src/wit-overview.md
```diff
@@ -0,0 +1,342 @@
...
+
+## Functions
+
+A function is defined by a name and a function type.  Like in record fields, the name is separated from the type by a colon:
+
+```
+do-nothing: func()
+```
+
+The function type is the word `func`, followed by a bracketed, comma-separated list of parameters (names and types). If the function returns a value, this is expressed as an arrow symbol (`->`) followed by the return type:
```
lann, line 218: I believe calling `()`s "brackets" is a British English thing; "brackets" typically means "square brackets" in the US.

A quick search suggests `()`s are most-unambiguously named "round brackets", though "parentheses" seems more common.

component-model/src/wit-overview.md
```diff
@@ -0,0 +1,342 @@
...
+```
+// This function does not return a value
+print: func(message: string)
+
+// These functions return values
+add: func(a: u64, b: u64) -> u64
+lookup: func(store: kv-store, key: string) -> option<string>
+```
+
+A function can have multiple return values; this is logically equivalent to returning a tuple but may be represented differently.  Refer to [the specification](https://github.com/WebAssembly/component-model/blob/main/design/mvp/WIT.md#wit-functions) if you're interested in this feature.
```
lann, line 229: > this is logically equivalent to returning a tuple but may be represented differently

I would maybe leave this off, given that the linked spec sections says:

> Note that returning multiple values from a function is not equivalent to returning a tuple of values from a function.

component-model/src/wit-overview.md
```diff
@@ -0,0 +1,342 @@
...
+    record point {
+        x: u32,
+        y: u32,
+    }
+
+    draw-line: func(canvas: canvas-id, from: point, to: point)
+}
+```
+
+Items in an interface should _not_ be separated by anything other than whitespace.
```
lann, line 250: ...and comments?

component-model/src/wit-overview.md
```diff
@@ -0,0 +1,342 @@
...
+}
+```
+
+> Even if you are only using one type, it must still be enclosed in braces. For example, `use types.{dimension}` is legal but `use types.dimension` is not.
+
+This works across files as long as the files are in the same package (effectively, in the same directory). For information about using definitions from other packages, see [the specification](https://github.com/WebAssembly/component-model/blob/main/design/mvp/WIT.md#interfaces-worlds-and-use).
+
+## Worlds
+
+A world describes a set of imports and exports, enclosed in braces and introduced with the `world` keyword. Roughly, a world describes the contract of a component. Exports are implemented by the component, and define what consumers of the component may call; imports are things the component may call. The imports and exports may be interfaces or individual functions.
```
lann, line 278: nit: I would say "provided" by the component, not "implemented"; components can reuse an import (implemented elsewhere) to fulfill an export.

component-model/src/wit-overview.md
```diff
@@ -0,0 +1,342 @@
...
+You can import and export interfaces defined in other packages. This can be done using `package/name` syntax:
+
+```
+world http-proxy {
+    export wasi:http/incoming-handler
+    import wasi:http/outgoing-handler
+}
+```
+
+_TODO: how does it resolve those package names?_
```
lann, line 310: Not specified by WIT; it's up to the tooling (a la crates and cargo). Could be e.g. a filesystem or a registry.

lann (commented):
I know this isn't finished, but reviewing helps me crystalize my own understanding too.

kate-goldenring:
@itowlson I think keeping it one page could work if we add a table of contents

component-model/src/wit-overview.md
itowlson: What I am trying to get at is that multiple return values is a product type in the same way as a tuple or record (for example, in Rust a MRV is surfaced as "return a tuple").

Maybe I should just leave MRV out of this overview - is it something that people are going to use/encounter ordinarily?  If it is common, in what way is it "not equivalent" to returning a tuple?  Is it to do with Go bindings?

lann: I _believe_ that `func() -> (a: u32, b: float32)` and `func() -> tuple<u32, float32>` are incompatible types.

lann: I guess the same reason(s) you would specify multiple _input_ parameters rather than one struct with multiple fields. :shrug:

itowlson:
@kate-goldenring I added a TOC but am not sure how I feel about it - please take a look!

lann:
Maybe something like https://github.com/badboy/mdbook-toc ?

kate-goldenring:
@itowlson i definitely like the TOC addition you made. If we can add an mdbook specific tool like the one @lann mentioned so it can automatically update that would be great

itowlson:
@kate-goldenring I have reservations about trying to wham everything in at the top, and until we know of an example that is both comprehensive, reusable through the book, _and_ cohesive, I'd be inclined to hold off on guessing.  Definitely worth exploring in a follow-up though!

itowlson:
Related: #9

itowlson:
@kate-goldenring I've squashed the commits. Any blockers to getting this in?

kate-goldenring (approved):

Issue #6, opened 2023-08-09
Create docs: Concepts


kate-goldenring:
Per [proposed TOC](https://github.com/bytecodealliance/component-docs/pull/1), add section that includes:

- General discussion of WIT as the IDL
- Interfaces
- Worlds
- Packages
- Capabilities
- Resources

_concepts build upon each other. example oriented rather than just a reference_
See Lann's cheatsheet for potential inspiration: https://lann.github.io/wit-quick-ref/

Issue #2, opened 2023-08-09
Create docs: Introduction to the Component Model


kate-goldenring:
Per [proposed TOC](https://github.com/bytecodealliance/component-docs/pull/1), add section that includes: 
- How to use this book - who is this for, how to use this book, how to contribute, background info/prerequisite knowledge
- Roadmap
  - Current Status
  - Future Directions
- Implementations - where can I run components (runtimes, languages, etc)

kate-goldenring:
closing this for a more specific issue https://github.com/bytecodealliance/component-docs/issues/105

PR #15, opened 2023-08-15
Concepts

itowlson:
This is a first crack at a basic concepts page, though it probably recapitulates the WIT stuff too much.  It was originally conceived of to cover things like resources and the capability model, but I think those are out of scope for Preview 2?

I was originally hoping to cover the "What is a Component?" topic as part of this, but after writing and deleting multiple attempts, I have concluded I still have _no idea_ what a component is!  This is going to be one of those monad things where as soon as I understand it, I become unable to explain it, isn't it...


itowlson:
I think for "what is a component" I probably need to stop trying to define what a component _is_, and instead take the reader on a journey from "what we can do with core modules like you already know about" through "what we _can't_ do with core modules even though it would be super handy" and thereby get to more "what components do for you."

Which seems obvious in retrospect but I wish it had seemed obvious a few hundred words earlier...

itowlson:
Well I am still not very happy with this but am stumped as to how to do better - very open to any feedback (even if it's "bin it all and write this instead").  Thanks!

itowlson:
Relates to #2 and #6.

component-model/src/concepts.md
```diff
@@ -0,0 +1,44 @@
...
+
+The component model defines a **canonical ABI** - an ABI to which all components adhere. This guarantees that components can talk to each other without confusion, even if they are built in different languages. Internally, a C component might represent strings in a quite different way from a Rust component, but the canonical ABI provides a format for them to pass strings across the boundary between them.
+
+## Worlds
+
+A **world** is a higher-level contract that describes a component's capabilities and needs.
+
+In one sense, a world _defines_ a component - it says what interfaces the component exposes for other code to call, and what interfaces the component depends on - but it defines only the surface of a component, not the internal behaviour. This is a useful way to think about "business logic" components, like libraries that export APIs for other components to use. Most of the worlds you define will be these kind of custom, library worlds.
```
kate-goldenring, line 22: i really like this explanation that it defines the 
"surface" of a component

component-model/src/concepts.md
```diff
@@ -0,0 +1,44 @@
...
+## Canonical ABI
+
+An ABI is an **application binary interface** - an agreement on how to pass data around in a binary format. ABIs are specifically concerned with data layout at the bits-and-bytes level. For example, an ABI might define how integers are represented (big-endian or little-endian?), how strings are represented (pointer to null-terminated character sequence or length-prefixd? UTF-8 or UTF-16 encoded?), and how composite types are represented (the offsets of each field from the start of the structure).
+
+The component model defines a **canonical ABI** - an ABI to which all components adhere. This guarantees that components can talk to each other without confusion, even if they are built in different languages. Internally, a C component might represent strings in a quite different way from a Rust component, but the canonical ABI provides a format for them to pass strings across the boundary between them.
+
+## Worlds
+
+A **world** is a higher-level contract that describes a component's capabilities and needs.
+
```
kate-goldenring, line 21: > Most of the worlds you define will be these kind of custom, library worlds.

By "define" are you emphasizing that the components that make up your business logic likely have custom worlds while those exposed to the runtime and host are likely common worlds (that likely live in `warg`)?

component-model/src/concepts.md
```diff
@@ -0,0 +1,44 @@
...
+A world is composed of interfaces, but each interface is _directional_ - it indicates whether the interface is available for outside code to call, or outside code must fulfil the interface for the component to call. These interfaces strictly bounds the component. A component cannot interact with anything outside itself except by having its exports called, or by calling its imports. This provides very strong sandboxing: for example, if a component does not have an import for a secret store, then it _cannot_ access that secret store, even if the store is running in the same process.
```
kate-goldenring, line 26: is it worth mentioning capability based security here or linking to it (external link or to a section we may create)?

component-model/src/concepts.md
```diff
@@ -0,0 +1,44 @@
...
+
+For a component to run, its imports must be fulfilled, by a host or by other components.  Connecting up one component's imports to another component's matching exports is called _composition_.
+
+* A (trivial) "HTTP proxy" world would export a "receive HTTP requests" interface, and import a "send HTTP requests" interface. A host, or another component, would call the exported "receive" interface, passing an HTTP request; the component would forward it on via the imported "send" interface. To be a _useful_ proxy, the component would also need to import interfaces such as I/O and clock time - without those imports the component could not perform, for example, caching.
+
+## Components
+
+Physically, a **component** is a specially formatted WebAssembly file. Internally, the component could include multiple traditional WebAssembly modules, and sub-components, composed via their imports and exports.
```
kate-goldenring, line 34:
```suggestion
Physically, a **component** is a specially formatted WebAssembly file. Internally, the component could include multiple traditional (or core) WebAssembly modules, and sub-components, composed via their imports and exports.
```

component-model/src/concepts.md
```diff
@@ -0,0 +1,44 @@
...
+
+The external interface of a component - its imports and exports - corresponds to a world. The component, however, defines behaviour and internal state.
+
+## Packages
+
+A **package** is a set of one or more WIT (Wasm Interface Type) files containing a related set of interfaces and worlds. WIT is an IDL (interface definition language) for Wasm. Packages provide a way for worlds and interfaces to refer to each other, and thus for an ecosystem of components to share common definitions.
+
+A package is not a world. It's a way of grouping related interfaces and worlds together for ease of discovery and reference, more like a namespace.
+
```
kate-goldenring, line 43: should we have this proxy be a part of our tutorial for continuity?

component-model/src/what-is-a-component.md
```diff
@@ -0,0 +1,15 @@
+# What is a Component?
+
+If you've tried out WebAssembly, you'll be familiar with the concept of a _module_. Roughly speaking, a module corresponds to a single `.wasm` file, with functions, types, memory, imports and exports, and so on. These modules can run in the browser, or via a separate runtime such as Wasmtime or Wamr. A module is what you get by default when you compile a program written in Rust, C, Go or whatever to WebAssembly.
```
kate-goldenring, line 3: What do you mean by "get by default"? The fact that right now when you specify a wasm target in those languages it compile to a module? Do we expect that to change to be directly to components in the future?

component-model/src/what-is-a-component.md
```diff
@@ -0,0 +1,15 @@
...
+
+Modules are, however, limited to describing themselves in terms of core WebAssembly types - integers and floating-point numbers. Just as in native assembly code, richer types, such as strings or records (structs), have to be represented in terms of integers and floating point numbers, for example by the use of pointers and offsets. And just as in native code, those representations are not interchangeable. A string in C might be represented entirely differently from a string in Rust, or a string in JavaScript.
```
kate-goldenring, line 5: Somewhere in here, i feel like we should use the term "core module" or remark that it is synonymous as i've heard it used in the ecosystem to distinguish components from modules.
```suggestion
Modules (often called "core modules") are, however, limited to describing themselves in terms of core WebAssembly types - integers and floating-point numbers. Just as in native assembly code, richer types, such as strings or records (structs), have to be represented in terms of integers and floating point numbers, for example by the use of pointers and offsets. And just as in native code, those representations are not interchangeable. A string in C might be represented entirely differently from a string in Rust, or a string in JavaScript.
```

kate-goldenring (approved):
I really like this! I think we can continue to add to concepts and expand upon it but overall this already unveils a lot of information.

component-model/src/concepts.md
itowlson: This was trying to refine Joel's observation that "most people don't define worlds unless they're building an operating system-a-like".  I'm trying to highlight that as far as _most readers_ are concerned, they'll only define their own worlds when creating APIs for libraries; for big, common worlds like WASI, they will _consume_ then but not _define_ them.  I'll see if I can phrase this more clearly but open to suggestions!

component-model/src/concepts.md
```diff
@@ -0,0 +1,44 @@
...
+## Canonical ABI
+
+An ABI is an **application binary interface** - an agreement on how to pass data around in a binary format. ABIs are specifically concerned with data layout at the bits-and-bytes level. For example, an ABI might define how integers are represented (big-endian or little-endian?), how strings are represented (pointer to null-terminated character sequence or length-prefixd? UTF-8 or UTF-16 encoded?), and how composite types are represented (the offsets of each field from the start of the structure).
+
+The component model defines a **canonical ABI** - an ABI to which all components adhere. This guarantees that components can talk to each other without confusion, even if they are built in different languages. Internally, a C component might represent strings in a quite different way from a Rust component, but the canonical ABI provides a format for them to pass strings across the boundary between them.
+
+## Worlds
+
+A **world** is a higher-level contract that describes a component's capabilities and needs.
```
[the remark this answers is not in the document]
itowlson: I think (and I realise that Dan is the authority here but YOLO) that's wrong though - I originally wrote in very much those terms, but "an environment" is super confusing when you're going "okay I am writing a regex parsing component" and its "world" is `export parse_regex: func(regex: string) -> parsed-regex`.  That's not an environment!  That's an API!  Dan is clearly writing about the WASI worlds, but the component model takes advantage of duality to use the same concept for _hosted things_ as well as hosts.  Which is jolly ingenious at a technical level but makes posts like Dan's rather misleading...

component-model/src/concepts.md
itowlson: I think that depends on handles and resources which don't exist yet, or am I misunderstanding?

component-model/src/concepts.md
itowlson: I feel like a semi-realistic HTTP proxy might be too complicated and a simplified one might be unconvincing.  Probably it would be better to design the tutorial examples for maximum, er, tutoring, and then cite them here, rather than the other way round.  But we can reassess over time.

component-model/src/what-is-a-component.md
itowlson: Yeah, that is what I was thinking.  I'd be kind of surprised if everything switched to components by default - core modules are the base Wasm set and I'd be sceptical about Chrome and Firefox adopting the component model - but I guess it does depend on what the compile target is.  I will reword.

component-model/src/concepts.md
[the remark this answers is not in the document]
kate-goldenring: Thats an interesting point that component to component worlds can look more api like while boundary components can have more environment like worlds. Thanks for sharing these thoughts. We might want to discuss "component behavior patterns" like this somewhere or just see if the patterns continue to unfold first.

component-model/src/concepts.md
kate-goldenring: > for big, common worlds like WASI, they will consume then but not define them.

I think this is the key. maybe we can just drop this line in

component-model/src/concepts.md
kate-goldenring: fair point. waiting makes sense.

component-model/src/concepts.md
kate-goldenring: the `wasm-tools compose` [example](https://github.com/bytecodealliance/wasm-tools/tree/main/crates/wasm-compose/example) is essentially a proxy (called `middleware`) and it isn't too hard to understand

kate-goldenring (approved):

component-model/src/concepts.md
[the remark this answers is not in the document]
sunfishcode: Fwiw, I like your use of "contract" in the text here, and what you're saying here makes sense to me.

I'd advise caution about thinking about worlds as "APIs" though, because they're not just "here are some functions users can call". They're also, "and here's what those functions themselves can call", which is more than what I think "API" is commonly understood to mean. For example, even though that `parse_regex` world looks like an API, it's an API plus a declaration that the implementation doesn't call any other APIs.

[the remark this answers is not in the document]
itowlson: @sunfishcode Yes, that's a good point - even a library world declares its dependencies / imports.  I do have a sentence for library-style components that "In one sense, a world _defines_ a component - it says what interfaces the component exposes for other code to call, and **what interfaces the component depends on**" (emphasis added) but it would do no harm to re-emphasise it where I talk about "custom worlds to expose ... functionality".  Thanks!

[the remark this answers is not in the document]
itowlson: @sunfishcode I added some wording, and a couple of examples, including your one about the `parse_regex` world declaring that it doesn't call any other APIs (which to me is a very powerful part of the component model).  As these are small changes, I won't take them round for another review cycle, but please flag up if you think they're mistaken or unclear.  Thanks again!

itowlson:
Merging this with a view to continuing to iterate on it as the book develops!

Issue #7, opened 2023-08-09
Create docs: Creating and Consuming Components


kate-goldenring:
Per [proposed TOC](https://github.com/bytecodealliance/component-docs/pull/1), add section that includes:
 
- how to author a component, combine components, call components, validate a component, distribute a component

PR #16, opened 2023-08-17
Creating and consuming components

itowlson:
This is really early at the moment, just creating the PR so the shape is visible (and in case anyone wants to weigh in).

### DIFF
--- a/component-model/src/SUMMARY.md
+++ b/component-model/src/SUMMARY.md
@@ -2,3 +2,7 @@
 
 - [Design](./design.md)
   - [An Overview of WIT](./wit-overview.md)
+- [Creating and Consuming Components](./creating-and-consuming-components.md)
+  - [Overview](./creating-and-consuming-components/overview.md)
+  - [Creating Components with Rust](./creating-and-consuming-components/creating-rust.md)
+  - [Consuming Components with Rust](./creating-and-consuming-components/consuming-rust.md)
--- a/component-model/src/creating-and-consuming-components.md
+++ b/component-model/src/creating-and-consuming-components.md
@@ -1,0 +1,1 @@
+# Creating and Consuming Components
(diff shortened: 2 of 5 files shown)


itowlson:
Relates to #7.

kate-goldenring:
@itowlson thanks for getting this started! A few thoughts:
1. this looks like it has some duplication to the language guide https://github.com/bytecodealliance/component-docs/pull/17. Should we point to the rust section of the language guide to kickstart them creating and calling a component and then pick up with the details on composing, validating, and distributing components? Or do we think we will need to repeat creating components in order to create multiple for compose examples?

2. If we do cover component creation in here, should we just create components using WAT and `wasm-tools` rather than a language toolchain?

3. In the TOC, we say that the goal of this document is to explain: how to author a component, combine components, call components, validate a component, distribute a component. Instead of a tutorial like approach, we could keep it generic in the examples: `wasm-tools compose -c config.yml -o component_one.wasm component_two.wasm`

itowlson:
@kate-goldenring Yes, maybe this is a duplicate of the language tools docs - sorry - I will rework.

>  should we just create components using WAT and wasm-tools

I reckon not, nobody wants to write WAT, and only the desperate want to read it.  _grin_  We can use it for explaining the internal structure of a Wasm file but I wouldn't make it a prominent way to do things.

> In the TOC, we say that the goal of this document is to explain: how to author a component, combine components, call components, validate a component, distribute a component. Instead of a tutorial like approach, we could keep it generic

Yeah, I think this was part of where I was trying to go with this doc, but I derailed myself into capturing the cargo component stuff (which as you rightly note is more a language thing).  There is probably some kerjiggering to do in the TOC once we have a better concept of the content - maybe "creating and consuming" is actually the language guide, and "combine, validate and distribute" is a different thing - or maybe the language guides sit alongside the language-agnostic tools under "creating and consuming" - not sure.  I guess maybe try to knock the content out and then see what organisation makes the most sense for it?

kate-goldenring:
> maybe "creating and consuming" is actually the language guide, and "combine, validate and distribute" is a different thing 

This seems to me where we are headed. I am happy with the content approach and then go from there

itowlson:
Replaced this with #19.

Issue #8, opened 2023-08-09
Create docs: language support and runtimes


kate-goldenring:
Per [proposed TOC](https://github.com/bytecodealliance/component-docs/pull/1), add section that includes:
Language toolchain and runtime overview for:
- C
- Go
- JavaScript
- Python
- Rust

Highlights the language toolchains that support components currently: namely `jco`, `cargo component` and `componentize py`

kate-goldenring:
Closed by https://github.com/bytecodealliance/component-docs/pull/17

PR #17, opened 2023-08-17
Language Support Guide for Components

kate-goldenring:
ref https://github.com/bytecodealliance/component-docs/issues/8


component-model/src/language-support.md
```diff
@@ -0,0 +1,301 @@
+# Wasm Language Support
+
+WebAssembly can be targeted by the majority of top programming
+languages; however, the level of
+support varies. This document details the subset of languages that target WASI and support
+components. This is a living document, so if you are aware of advancements in a toolchain, please do
+not hesitate to [contribute documentation](contributing.md). Each section covers how to build and
```
kate-goldenring, line 7: non-existant currently :)

component-model/src/language-support.md
```diff
@@ -0,0 +1,301 @@
...
+One of the benefits of components is their portability across host runtimes. The runtime only needs
+to know what world the component is targeting in order to import or execute the component. This
+language guide hopes to demonstrate that with a prevailing `example` world defined in
+[`../examples/add.wit`](../examples/add.wit). Furthermore, an example host that understands the `example`
+world has been provided in [`../examples/add-host`](../examples/add-host/) for running components. Each
+toolchain section walks through creating a component of this world, which can be run either in the
+example host or from an application of that toolchain. This aims to provide a full story for using
+components within and among toolchains.
+ 
+- [Wasm Language Support](#wasm-language-support)
```
itowlson, line 19: Maybe too much detail at this stage but I'd love to see these broken out into separate pages so readers could home in on them (and flip between them) via the main TOC.

component-model/src/language-support.md
```diff
@@ -0,0 +1,301 @@
...
+
+## JavaScript Tooling
+
+[`jco`](https://github.com/bytecodealliance/jco) is a fully native JS tool for working with the
+emerging WebAssembly Components specification in JavaScript.
+
+### Building a Component with `jco`
+
+A component can be created from a JS module using `jco componentize`. First, install `jco` and
+`componetize-js`:
```
itowlson, line 147:
```suggestion
`componentize-js`:
```

component-model/src/language-support.md
```diff
@@ -0,0 +1,301 @@
...
+Now, use `jco` to create a component from the JS module:
+
+```sh
+$ jco componentize add.js --wit add.wit -n example -o add.wasm
+OK Successfully written add.wasm with imports ().
+```
+
+Now, run the component using the Rust `add` host:
+
+> Note: it can take over 20 seconds to complete
```
itowlson, line 171: HOLY CRAP

component-model/src/language-support.md
```diff
@@ -0,0 +1,301 @@
...
+```
+
+A core module and JavaScript bindings have been outputted to the `out-dir`.
+
+Now, you can import the resultant `add.js` file and run it from a JavaScript application. This
+example imports it as an ECMAScript module for ease of running locally with node:
+
+```mjs
+// app.mjs
+import { add } from './out-dir/add.mjs';
```
itowlson, line 201: there isn't an `add.mjs` listed above?

_ETA: oh it gets renamed later?_

component-model/src/language-support.md
```diff
@@ -0,0 +1,301 @@
...
+
+```sh
+$ pip install componentize-py
+$ componentize-py -d ../examples/add.wit -w example componentize guest -o add.wasm 
+Component built successfully
+```
+
+To test the component, run it using the Rust `add` host:
+
+> Note: it can take over 30 seconds to complete
```
itowlson, line 246: I regret using up HOLY CRAP on the 20-second one

itowlson (approved):
Main suggestion is to break this into language pages but it's not a blocker.  Thanks, looks good!

component-model/src/language-support.md
kate-goldenring: I've misspelled this so many times. writing the word c-o-m-p-o-n-e-n-t-i-z-e launches my brain to another realm

component-model/src/language-support.md
kate-goldenring: @dicej can you provide some context on why it takes so long to run a JS or Python component from a Rust host? I am assuming it is initialization costs that could be cut down if the component is pre-initialized with Wizer which might be helpful to add to this note. Maybe later on we can add a section with all the ways to optimize running a component

component-model/src/language-support.md
kate-goldenring: added a couple words ("and rename") to make this less of a surprise

component-model/src/language-support.md
dicej: It definitely shouldn't take that long.  Are you building the host using `--release`?

`componentize-py` already does pre-initialization using [component-init](https://github.com/dicej/component-init), which is equivalent to Wizer but for components rather than core modules.

dicej: Oh, yeah, I see now.  Use `cargo run --release ...`.

dicej: And note that Wasmtime has to compile the .wasm to a .cwasm every time.  In a performance-sensitive environment, we would enable .cwasm caching and/or pre-instantiate, bringing the performance down into the sub-millisecond range.  See also https://github.com/dicej/wasmtime-serverless-performance

kate-goldenring: that did the trick. Thanks Joel. I have a naive view on what the `release` `opt-level` does in Rust where i assumed most optimizations were on reducing the binary size. This completely shortens the wait time -- do you have a sense of what some of the major optimizations that are happening here?

component-model/src/language-support.md
dicej: Only one HOLY CRAP per code review, @itowlson.  You know the rules.

component-model/src/language-support.md
dicej: The machine code generated by LLVM with optimization disabled is really quite terrible from a performance standpoint, especially for CPU-intensive tasks like what Cranelift has to do to transpile a large amount of Wasm (e.g. CPython or SpiderMonkey) to native code.  It's not usually noticeable for non-CPU intensive apps, but Wasmtime+Cranelift do a lot of work and benefit immensely from optimization, as you discovered.

component-model/src/language-support.md
kate-goldenring: i can break this out in a subsequent PR

component-model/src/language-support.md
itowlson: I existed it! (though with an all-caps name)

component-model/src/language-support.md
```diff
@@ -0,0 +1,296 @@
...
+`componentize-py` pulls in extra WASI Preview 2 imports, even if they are not used by the component.
+Currently, language toolchains are likely to pull in more than a component declares in WAT.
+
+```py
+from add import Root, RootImports
+from wasmtime import Store
+
+def main():
+    store = Store()
+    component = Root(store, RootImports(language=Host(), poll=None, monotonic_clock=None, wall_clock=None, streams=None, filesystem=None, random=None, environment=None, preopens=None, exit=None, stdin=None, stdout=None, stderr=None))
```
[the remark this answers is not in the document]
kate-goldenring: Thank you for catching this. This was from a previous example where the component imported a host `print` function. Fixed

PR #20, opened 2023-08-22
Publish component model book

itowlson:
This would use Fermyon Cloud as the publishing platform - I'm not trying to force this on anyone it's just the one I know how to deploy to.

The publish workflow needs a Fermyon Cloud access token - I don't have permission to add secrets to the repo, so if this is accepted then someone else will need to create and add one.

### DIFF
--- a/.github/workflows/component-model.yml
+++ b/.github/workflows/component-model.yml
@@ -1,0 +1,34 @@
+name: Publish Component Model book
+on:
+  push:
+    paths:
+      - component-model/**
+    branches:
+      - main
+  workflow_dispatch:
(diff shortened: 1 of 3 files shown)


itowlson:
I've set up a publishing token - it's currently for my FC account but longer term should go to an org account (if the org decides to stay with FC, otherwise this will be superseded anyway).

kate-goldenring (approved):
@itowlson this is a great way to get this up and accessible. Thanks for adding this

itowlson:
https://component-model-book.fermyon.app/

PR #23, opened 2023-08-24
Front matter

itowlson:
Brief introduction and a slight tweak to the TOC.

### DIFF
--- a/component-model/src/introduction.md
+++ b/component-model/src/introduction.md
@@ -1,0 +1,25 @@
+# Introduction
+
+This book introduces the WebAssembly Component Model, a broad-reaching architecture for building interoperable Wasm libraries, applications, and environments.
+
+## Who is this book for?
+
+This book is aimed at _users_ of the component model - developers building libraries and applications. It covers how to build Wasm components, and how to use Wasm components to build applications or higher-level libraries.
+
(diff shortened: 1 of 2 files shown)


component-model/src/introduction.md
```diff
@@ -0,0 +1,25 @@
...
+
+## Who is this book for?
+
+This book is aimed at _users_ of the component model - developers building libraries and applications. It covers how to build Wasm components, and how to use Wasm components to build applications or higher-level libraries.
+
+The book does not cover the _implementation_ of the component model, such as building Wasm runtimes or hosts.
+
+## Status
+
+The component model is a work in progress. Although the architecture is well-defined, details are still evolving. This version of the guide is current as at late August 2023.
```
kate-goldenring, line 13: i get worried about putting a date here and forgetting to update it. Is there a way we can be sure to update it? Maybe a github action?

kate-goldenring (approved):

component-model/src/introduction.md
itowlson: Not sure... I'd be wary of using an action anyway - e.g. if we fix a typo but not the tool docs then would an action update the date incorrectly?

kate-goldenring: Yeah maybe we just try to check it at the end of the month.

PR #28, opened 2023-08-24
"What is a Component?" edits

lann:


component-model/src/what-is-a-component.md
```diff
@@ -1,17 +1,17 @@
 # What is a Component?
 
-If you've tried out WebAssembly, you'll be familiar with the concept of a _module_. Roughly speaking, a module corresponds to a single `.wasm` file, with functions, types, memory, imports and exports, and so on. These "core" modules can run in the browser, or via a separate runtime such as Wasmtime or Wamr. A module is defined by the base Wasm specification, and if you compile a program written in Rust, C, Go or whatever for the browser, then a core module is what you'd get.
+If you've tried out WebAssembly, you'll be familiar with the concept of a _module_. Roughly speaking, a module corresponds to a single `.wasm` file, with functions, memory, imports and exports, and so on. These "core" modules can run in the browser, or via a separate runtime such as Wasmtime or WAMR. A module is defined by the [WebAssembly Core Specification](https://webassembly.github.io/spec/core/), and if you compile a program written in Rust, C, Go or whatever for the browser, then a core module is what you'll get.
```
lann, line 3: Core modules don't have "types" in the same sense as components. They do have other things you could call "types" but might be better to leave those as part of "and so on" to reduce potential confusion.

component-model/src/what-is-a-component.md
```diff
@@ -1,17 +1,17 @@
...
 
-Core modules are, however, limited to describing themselves in terms of core WebAssembly types - integers and floating-point numbers. Just as in native assembly code, richer types, such as strings or records (structs), have to be represented in terms of integers and floating point numbers, for example by the use of pointers and offsets. And just as in native code, those representations are not interchangeable. A string in C might be represented entirely differently from a string in Rust, or a string in JavaScript.
+Core modules are, however, limited to describing themselves in terms of a small number of core WebAssembly types such as integers and floating-point numbers. Just as in native assembly code, richer types, such as strings or records (structs), have to be represented in terms of integers and floating point numbers, for example by the use of pointers and offsets. And just as in native code, those representations are not interchangeable. A string in C might be represented entirely differently from a string in Rust, or a string in JavaScript.
```
lann, line 5: "Just ints and floats" gets the right idea across but technically there are also reftypes.

component-model/src/what-is-a-component.md
```diff
@@ -1,17 +1,17 @@
...
 
-This guide doesn't try to formally specify what a component _is_. Physically, components are Wasm modules, such as `.wasm` files, with a specific internal format. Logically, components are containers for modules - or other components - which express their interfaces and dependencies via WIT and the Canonical ABI. Conceptually, components are self-describing units of code that interact only through interfaces instead of shared memory.
+This guide doesn't try to formally specify what a component _is_. Physically, components are Wasm data, such as `.wasm` files, with a specific internal format. Logically, components are containers for modules - or other components - which express their interfaces and dependencies via WIT and the Canonical ABI. Conceptually, components are self-describing units of code that interact only through interfaces instead of shared memory.
```
lann, line 17: I'm not really happy with "data" here but less unhappy than "components are Wasm modules" :smile:.

itowlson (approved):

PR #29, opened 2023-08-24
"An Overview of WIT" edits

lann:


component-model/src/wit-overview.md
```diff
@@ -235,7 +235,7 @@ flags allowed-methods {
 
 ### Type aliases
 
-You can define convenience names for types using `type ... = ...`. This can be useful for giving shorter or more meaningful names to types:
+You can define a new named type using `type ... = ...`. This can be useful for giving shorter or more meaningful names to types:
```
lann, line 238: There are some rules around types that _must_ be named if imported/exported that I haven't entirely wrapped my head around yet, but unfortunately I think it means that these aren't _just_ for convenience.

itowlson (approved):
wow, apparently I really like typing `im`

PR #32, opened 2023-08-31
Fix wrong item numbers

sachaos:
I've fixed the wrong item numbers in component-model/src/language-support/rust.md .


kate-goldenring (approved):
Thank you!

PR #56, opened 2023-09-08
Refactor and new index page

merlijn-sebrechts:
This updates the index page to

* add a table linking to different subsections of the docs
* point runtime devs to the spec docs

And refactors the structure to

* Put each concept on a separate page
* Turn "components" into "why the component model"

Apart from the index page, I haven't actually updated any text. This is by design; we can update the individual pages in separate PRs, but with this PR it's easy for other people to fix separate pages.


component-model/src/design/why-component-model.md
```diff
@@ -1,4 +1,4 @@
-# What is a Component?
+# Why the Component Model?
```
kate-goldenring, line 2: This paragraph looks like it could exist in `components.md`:

_This guide doesn't try to formally specify what a component is. Physically, components are Wasm data, such as .wasm files, with a specific internal format. Logically, components are containers for modules - or other components - which express their interfaces and dependencies via WIT and the Canonical ABI. Conceptually, components are self-describing units of code that interact only through interfaces instead of shared memory._

merlijn-sebrechts: Thanks, I moved it and added a link to the spec!

itowlson (approved):
This is great!  I really love the extra structure it provides, and the navigation from the home page is going to make users way more successful!

PR #61, opened 2023-09-08
Added categories and removed unused pages

merlijn-sebrechts:
So the sidebar looks like this:

> image


kate-goldenring:
I think i have a slight preference for a top level page for sections that state the purpose and possibly a summary of a section. For example, this update to design: https://github.com/bytecodealliance/component-docs/pull/64/files

kate-goldenring (approved):

PR #76, opened 2023-09-21
Generate the sitemap in CI

itowlson:
This builds on the static SEO work to generate the sitemap at deployment time instead of maintaining it by hand.


kate-goldenring (approved):
🎉

itowlson:
Tested the deployed site and it looks plausible

Issue #37, opened 2023-09-07
Language Guide: Add (tiny)go


kate-goldenring:
Add instructions for compiling go modules and componentizing them. For example:
```sh
tinygo build -target=wasi foo.go
wasm-tools component embed --world myworld ./foo.wit foo.wasm -o foo.embed.wasm
wasm-tools component new -o foo.component.wasm --adapt wasi_snapshot_preview1="$COMPONENT_ADAPTER" foo.embed.wasm
```
where the adapter is built from https://github.com/bytecodealliance/wasmtime/tree/main/crates/wasi-preview1-component-adapter

Mossaka:
I can take a look

kate-goldenring:
@Mossaka i started working on this here https://github.com/bytecodealliance/component-docs/pull/77. would you be willing to give it a review and if its really out of shape we can pivot

PR #77, opened 2023-09-23
Add Go language guide

kate-goldenring:
Add go language guide and a description of reactor vs command components as a lot of tooling refers to the two but we haven't explicitly called out the difference anywhere

fixes https://github.com/bytecodealliance/component-docs/issues/37

### DIFF
--- a/component-model/src/language-support/go.md
+++ b/component-model/src/language-support/go.md
@@ -13,13 +13,29 @@ There are several steps to building a component in TinyGo:
 1. Determine which world the component will implement
 2. Build a Wasm core module using the native TinyGo toolchain
 3. Convert the Wasm core module to a component using
-   []`wasm-tools`](https://github.com/bytecodealliance/wasm-tools)
+   [`wasm-tools`](https://github.com/bytecodealliance/wasm-tools)
 
-The next section will walk through steps 1-2, producing a core Wasm module that targets WASI preview 1.
-Then, the following section will walk through converting this core module to a component that
-supports WASI preview 2.
+The following sections will walk through these steps, producing a core Wasm module that targets WASI
+preview 1 and converting this core module to a component that supports WASI preview 2.
 
-## Creating a TinyGo Core Wasm Module
+### 1: The `example` World
+
+The next two sections walk through creating a component that implements the the following [`example`
+world](https://github.com/bytecodealliance/component-docs/tree/main/component-model/examples/example-host/add.wit):
+
+```wit
+package example:component;
+
+world example {
+    export add: func(x: s32, y: s32) -> s32;
+}
+```
+
+This is a simple world that exports one `add` function. If you want to go beyond a quick start to a
+more realistic example, jump to the [section on implementing worlds with
+interfaces](#implementing-worlds-with-interfaces-with-tinygo-and-wit-bindgen).
+
+### 2: Creating a TinyGo Core Wasm Module
 
 The TinyGo toolchain natively supports compiling Go programs to core Wasm modules. Let's create one that implements the `add` function in the [`example`
 world](https://github.com/bytecodealliance/component-docs/tree/main/component-model/examples/example-host/add.wit).
@@ -51,7 +67,7 @@ tinygo build -o add.wasm -target=wasi add.go
 
 You should now have an `add.wasm` module. But at the moment, this is a core module. In the next section, we will convert it into a component.
 
-## Converting a Wasm Core Module to a Component
+### 3: Converting a Wasm Core Module to a Component
 
 In the previous step, we produced a core module that implements our `example` world. We now want to
 convert to a component to gain the benefits of the component model, such as the ability to compose
@@ -107,7 +123,7 @@ world root {
 }
 ```
 
-## Testing an `add` Component
+### Testing an `add` Component
 
 To run our add component, we need to use a host program with a WASI runtime that understands the
 `example` world. We've provided an [`example-host`](../../examples/example-host/README.md) to do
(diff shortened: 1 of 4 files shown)


kate-goldenring:
Currently debugging why the resultant `add.component.wasm` is not compatible with our `example-host`. I am getting:
```
Error: Failed to instantiate the example world

Caused by:
    0: import `wasi:io/streams` has the wrong type
    1: instance export `check-write` has the wrong type
    2: expected func found nothing
```

kate-goldenring:
updating wasmtime version did the trick so this is ready for review

component-model/examples/example-host/Cargo.toml
```diff
@@ -6,7 +6,7 @@ edition = "2021"
 [dependencies]
 async-std = { version = "1.12.0", features = ["attributes"] }
 clap = { version = "4.3.19", features = ["derive"] }
-wasmtime = { git = "https://github.com/bytecodealliance/wasmtime", rev = "7b9189b", features = ["component-model"] }
-wasmtime-wasi = { git = "https://github.com/bytecodealliance/wasmtime", rev = "7b9189b" }
-wasi-cap-std-sync = { git = "https://github.com/bytecodealliance/wasmtime", rev = "7b9189b" }
+wasmtime = { git = "https://github.com/bytecodealliance/wasmtime", rev = "aec4b25", features = ["component-model"] }
```
itowlson, line 9: Do any of the other tutorials need updating to remain in line with the new build, or do they continue to work?

component-model/src/language-support/go.md
```diff
@@ -0,0 +1,89 @@
...
+
+> Note: to resolve the local path to import the bindings from `gen`, update `go.mod`:
+> ```sh
+> echo "replace github.com/yourusername/yourproject => /Path/to/add" >> go.mod
+> ```
+
+```go
+package main
+
+//go:generate wit-bindgen tiny-go ./add.wit -w example --out-dir=gen
```
itowlson, line 31: This looks like the line we just ran.  What is its role here?

component-model/src/language-support/go.md
```diff
@@ -0,0 +1,89 @@
...
+
+```sh
+wit-bindgen tiny-go ./add.wit -w example --out-dir=gen
+```
+
+Now, we can implement the generated `SetExample` type (in `gen/example.go`) that is exported by our component:
+
+> Note: to resolve the local path to import the bindings from `gen`, update `go.mod`:
+> ```sh
+> echo "replace github.com/yourusername/yourproject => /Path/to/add" >> go.mod
```
itowlson, line 25: What does `/Path/to/add` mean here?  Is it the WIT file?

itowlson (changes requested):
Thanks for doing this!  The spine of it is great, but reading it I sometimes felt a bit adrift.  Key suggestions would be:

1. A bit more of a roadmap up front. We can't go directly from busting out some Go to having a component, so here are the waypoints we're going to hit.
2. A bit more what and why, around things like the `init`, `SetExample` stuff.

But the core story is great!  So good to have this in here.

component-model/src/language-support/go.md
```diff
@@ -0,0 +1,89 @@
+# Go Tooling
```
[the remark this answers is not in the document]
kate-goldenring: I think a full book rev makes sense. I think it is better than leaving at "latest dev" since that already implies Wasmtime 13 changes

component-model/src/language-support/go.md
kate-goldenring: This makes it so you can automatically regenerate the bindings on builds if the WIT changes but it probably over complicates. I'll remove

component-model/src/language-support/go.md
```diff
@@ -0,0 +1,89 @@
...
+import (
+    . "github.com/yourusername/yourproject/gen"
+)
+
+type AddImpl struct {
+}
+
+func init() {
+    a := AddImpl{}
+    SetExample(a)
```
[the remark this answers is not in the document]
Mossaka: The reasoning behind using `init()` is to register **user implementation** as exported interfaces / functions.

kate-goldenring:
@itowlson do you mind giving this another review? I broke the guide up into two sections: 
1. a quickstart that just exports an add function
2. one that uses wit-bindgen to support more complex worlds
I also call out how a lot of the `wasm-tools` commands to componentize with be unnecessary once the `wasip2` target is added.

component-model/src/language-support/go.md
```diff
@@ -0,0 +1,272 @@
...
+package](../../examples/example-host/add.wit). The component will implement a simple add function.
+
+## Overview of Building a Component with TinyGo
+
+There are several steps to building a component in TinyGo:
+
+1. Determine which world the component will implement
+2. Build a Wasm core module using the native TinyGo toolchain
+3. Convert the Wasm core module to a component using
+   []`wasm-tools`](https://github.com/bytecodealliance/wasm-tools)
```
itowlson, line 16: Is the extra `]` before `wasm-tools` intentional?

component-model/src/language-support/go.md
```diff
@@ -0,0 +1,272 @@
...
+
+The next section will walk through steps 1-2, producing a core Wasm module that targets WASI preview 1.
```
itowlson, line 18: Should we line up the ordering here - this next section is actually step 2 not step 1 I think?

component-model/src/language-support/go.md
```diff
@@ -0,0 +1,272 @@
...
+The [`example`
+world](https://github.com/bytecodealliance/component-docs/tree/main/component-model/examples/example-host/add.wit) we were using in the previous sections simply exports a function. However, to use your component from another component, it must export an interface. This means we will need to use a tool to generate bindings to use as glue code, and adds a couple more steps (2-3) to building Wasm components with TinyGo:
+
+1. Determine which world the component will implement
+2. Generate bindings for that world using
+   [`wit-bindgen`](https://github.com/bytecodealliance/wit-bindgen#creating-a-component)
+3. Implement the interface defined in the bindings
+4. Build a Wasm core module using the native TinyGo toolchain
+5. Convert the Wasm core module to a component using
+   []`wasm-tools`](https://github.com/bytecodealliance/wasm-tools)
```
itowlson, line 134: Again just sanity-checking the extra `]`

itowlson (approved):
Couple of things to sanity-check but looks good!

component-model/src/language-support/go.md
kate-goldenring: I updated the guide to number the headings and added a section introducing the `example` world. What do you think? [2be771c](https://github.com/bytecodealliance/component-docs/pull/77/commits/2be771c1c92c0d827d97a5e59223e19d9453588b)

itowlson (approved):
This is great - thank you Kate!

PR #88, opened 2023-11-16
Update cargo component install doc

mh4ck-Thales:
The docs for installing the `cargo component` command was using the git repo and stating that there is not `crates.io` distribution of `cargo component`, which is no longer true (see [here](https://crates.io/crates/cargo-component)). This PR fixes that by documenting installing directly from `crates.io` and updating the info message.


itowlson (approved):
Thanks!

PR #100, opened 2024-02-28
Add HTML and node.js demo to the tutorial

NukeManDan:
I have a minimal demo here to illustrate that you can generate JavaScript bindings via this method, and a browser and node targeting application using the same core logic of the tutorial.

Would really appreciate any suggestions here:
- Am I using the right tools and patterns?
- Would it be beneficial to merge a more polished version of this into the main book here?

Thanks so much for the tooling and the great docs you have already! :grin: (feel free to close this PR if not useful/off-target for you)


component-model/examples/tutorial/www/README.md
kate-goldenring, line 1: Can we rename this `www` folder to be `jco`?  In a subsequent PR, we can put all the rust example implementations under a `wasmtime` folder so each of these are example implementations for different runtimes

component-model/examples/tutorial/www/README.md
```diff
@@ -0,0 +1,24 @@
+# Browser & Node Demo
+
+## Calculator with `jco`
+
+Using [`jco`](https://github.com/bytecodealliance/jco/blob/main/docs/src/example.md)
+See the [`jco` example](https://github.com/bytecodealliance/jco/blob/main/docs/src/example.md) that was used to create this minimal demo.
```
kate-goldenring, line 6:
```suggestion
This is an example implementation of running a component that exports the `calculate` interface from a JavaScript application. It uses`jco` to generate JavaScript bindings and shows how the same component can be executed in the browser or locally with Node. For another example of using `jco` with components in multiple environments, see the [`jco` example](https://github.com/bytecodealliance/jco/blob/main/docs/src/example.md)
Using [`jco`](https://github.com/bytecodealliance/jco/blob/main/docs/src/example.md)
See the [`jco` example](https://github.com/bytecodealliance/jco/blob/main/docs/src/example.md) that was used to create this minimal demo.
```

component-model/examples/tutorial/www/README.md
```diff
@@ -0,0 +1,24 @@
...
+
+```sh
+(cd calculator && cargo component build --release)
+(cd adder && cargo component build --release)
+(cd command && cargo component build --release)
+wasm-tools compose calculator/target/wasm32-wasi/release/calculator.wasm -d adder/target/wasm32-wasi/release/adder.wasm -o composed.wasm
+wasm-tools compose command/target/wasm32-wasi/release/command.wasm -d composed.wasm -o command.wasm
```
kate-goldenring, line 13: I believe we don't need the create `command.wasm` in this example since we are not executing it with a host component that expects the command world (as `wasmtime run` does)
```suggestion
```

kate-goldenring (commented):
Sorry for the delayed review. Thank you for putting this together. I just have a couple suggestions for providing the reader with some more context. Overall, this is a wonderful addition.

component-model/examples/tutorial/www/README.md
NukeManDan: Yeah, here I left exactly what is in the top level readme. But perhaps we would instead insist that `composed.wasm` exists from the top level readme build? As to keep things in DRY compliance :wink:

NukeManDan: [bded978](https://github.com/bytecodealliance/component-docs/pull/100/commits/bded97810d29a8392de24a2bf8a372386f4638a3) now has what I hope might be better, this assumes the commands are run from the `jco` dir, where the README for it is. Not sure if that is ideal... :thinking: Depends on what you want to do for multiple runtime examples using the `composed.wasm` consumed by other `Guest` languages.

kate-goldenring: > this assumes the commands are run from the jco dir

I think that its a fair assumption that they are run from the same dir as the README. In a subsequent PR, i can update the parent README to point to this one and welcome other guest lang examples.

kate-goldenring (approved):

PR #103, opened 2024-03-13
wit: add resources

### LINKED ISSUE #97: Add Resources to WIT Overview
Resources were added to WIT in July: https://github.com/WebAssembly/component-model/blob/main/design/mvp/WIT.md

We should add them to the WIT overview (and pull in any other updates): https://github.com/bytecodealliance/component-docs/blob/main/component-model/src/design/wit.md

kate-goldenring:
Mainly copied over from: https://github.com/WebAssembly/component-model/blob/main/design/mvp/WIT.md#item-resource

Also added a handles subsection

closes #97

### DIFF
--- a/component-model/src/design/wit.md
+++ b/component-model/src/design/wit.md
@@ -18,7 +18,6 @@ The WIT (Wasm Interface Type) language is used to define Component Model interfa
     - [Variants](#variants)
     - [Enums](#enums)
     - [Resources](#resources)
-      - [Handles](#handles)
     - [Flags](#flags)
     - [Type aliases](#type-aliases)
   - [Functions](#functions)
@@ -206,68 +205,54 @@ This can provide a simpler representation in languages without discriminated uni
 
 ### Resources
 
-A resource type defines a new abstract type an entity with a lifetime. Unlike
-records, they describe things that can't or shouldn't be copied by value.
-Instead their ownership can be passed between two components via a
-[handle](#handles). Resources are similar to a a `class` in many programming
-languages and can have constructors and methods.
+Resources are handles to some entity that lives outside of the component. They
+describe things that can't or shouldn't be copied by value; instead, their
+ownership or reference can be passed between two components via a handle. Unlike
+other WIT types which are simply plain data, resources only expose behavior
+through methods. Resources are _objects that implement_ an interface.
 
-For example, the following Wit defines a resource type and a function that takes
-and returns a handle to a `blob`:
-```wit
-resource blob;
-transform: func(blob) -> blob;
-```
-
-As syntactic sugar, resource statements can also declare any number of
-*methods*, which are functions that implicitly take a `self` parameter that is a
-handle. A resource statement can also contain any number of *static functions*,
-which do not have an implicit `self` parameter but are meant to be lexically
-nested in the scope of the resource type. Lastly, a resource statement can
-contain at most one *constructor* function, which is syntactic sugar for a
-function returning a handle of the containing resource type.
-
-For example, the following is a resource with one constructor, two methods, and
-a static function:
+For example, we could model a blob (binary large object) as a resource. The
+following WIT defines the a `blob` resource type, which contains a constructor,
+two methods, and a static function:
 
 ```wit
 resource blob {
     constructor(init: list<u8>);
     write: func(bytes: list<u8>);
     read: func(n: u32) -> list<u8>;
-    merge: static func(lhs: borrow<blob>, rhs: borrow<blob>) -> blob;
+    merge: static func(lhs: blob, rhs: blob) -> blob;
 }
 ```
 
-Methods always desugar to a borrowed self parameter whereas constructors always
-desugar to an owned return value. For example, the `blob` constructor
-[above](#resources) desugars to:
+As shown in the `blob` example, a resource can contain:
+
+- _methods_: functions that implicitly take a `self` (often called `this` in many languages)
+parameter that is a handle
+- _static functions_: functions which do not have an implicit `self` parameter
+but are meant to be nested in the scope of the resource type
+- at most one _constructor_: a function that is syntactic sugar for a function
+returning a handle of the containing resource type
+
+Methods always desugar to a borrowed `self` parameter whereas constructors
+always desugar to an owned return value. For example, the `blob` resource
+[above](#resources) could be approximated as:
 
 ```wit
-resource blob;
-%[constructor]blob: func(self: borrow<blob>, bytes: list<u8>) -> blob;
-%[method]blob.write: func(self: borrow<blob>, bytes: list<u8>);
-%[method]blob.read: func(self: borrow<blob>, n: u32) -> list<u8>;
-%[static]blob.merge: func(lhs: borrow<blob>, rhs: borrow<blob>) -> blob;
+resource blob;  
+blob-constructor: func(bytes: list<u8>) -> blob;  
+blob-write: func(self: borrow<blob>, bytes: list<u8>);  
+blob-read: func(self: borrow<blob>, n: u32) -> list<u8>;  
+blob-merge: static func(lhs: blob, rhs: blob) -> blob;
 ```
 
-#### Handles
-
-There are two types of handles in Wit: "owned" handles and "borrowed" handles.
-Owned handles represent the passing of unique ownership of a resource between
-two components. When the owner of an owned handle drops that handle, the
-resource is destroyed. In contrast, a borrowed handle represents a temporary
-loan of a handle from the caller to the callee for the duration of the call.
-
-When a resource type name is used directly (e.g. when `blob` is used as the
-return value of the `blob` constructor [above](#resources)), it stands for an
-"owning" handle that will call the resource's destructor when dropped. When a
-resource type name is wrapped with `borrow<...>`, it stands for a "borrowed"
-handle that will *not* call the destructor when dropped.
+> A note on resource lifetimes: When the owner of an owned resource drops that
+resource, the resource is destroyed. In contrast, a borrowed resource represents
+a temporary loan of a resource from the caller to the callee for the duration of
+the call.
 
 ### Flags
 
-A flag type is a set of named booleans.  In an instance of the type, each flag will be either true or false.
+A flags type is a set of named booleans.  In an instance of the type, each flag will be either true or false.
 
 ```wit
 flags allowed-methods {
--- a/component-model/src/design/wit.md
+++ b/component-model/src/design/wit.md
@@ -1,6 +1,6 @@
 # An Overview of WIT
 
-The WIT (Wasm Interface Type) language is used to define Component Model interfaces and worlds. WIT isn't a general-purpose coding language and doesn't define behaviour; it defines only _contracts_ between components. This topic provides an overview of key elements of the WIT language. The official WIT specification and history can be found in the [`WebAssembly/component-model` repository](https://github.com/WebAssembly/component-model/blob/main/design/mvp/WIT.md).
+The WIT (Wasm Interface Type) language is used to define Component Model interfaces and [worlds](#worlds). WIT isn't a general-purpose coding language and doesn't define behaviour; it defines only _contracts_ between components. This topic provides an overview of key elements of the WIT language. The official WIT specification and history can be found in the [`WebAssembly/component-model` repository](https://github.com/WebAssembly/component-model/blob/main/design/mvp/WIT.md).
 
 - [An Overview of WIT](#an-overview-of-wit)
   - [Structure of a WIT file](#structure-of-a-wit-file)
--- a/component-model/src/design/wit.md
+++ b/component-model/src/design/wit.md
@@ -1,6 +1,6 @@
 # An Overview of WIT
 
-The WIT (Wasm Interface Type) language is used to define Component Model interfaces and worlds. WIT isn't a general-purpose coding language and doesn't define behaviour; it defines only _contracts_ between components. This topic provides an overview of key elements of the WIT language.
+The WIT (Wasm Interface Type) language is used to define Component Model interfaces and worlds. WIT isn't a general-purpose coding language and doesn't define behaviour; it defines only _contracts_ between components. This topic provides an overview of key elements of the WIT language. The official WIT specification and history can be found in the [`WebAssembly/component-model` repository](https://github.com/WebAssembly/component-model/blob/main/design/mvp/WIT.md).
 
 - [An Overview of WIT](#an-overview-of-wit)
   - [Structure of a WIT file](#structure-of-a-wit-file)
@@ -17,6 +17,8 @@ The WIT (Wasm Interface Type) language is used to define Component Model interfa
     - [Records](#records)
     - [Variants](#variants)
     - [Enums](#enums)
+    - [Resources](#resources)
+      - [Handles](#handles)
     - [Flags](#flags)
     - [Type aliases](#type-aliases)
   - [Functions](#functions)
@@ -202,9 +204,70 @@ enum color {
 
 This can provide a simpler representation in languages without discriminated unions. For example, a WIT enum can translate directly to a C++ `enum`.
 
+### Resources
+
+A resource type defines a new abstract type an entity with a lifetime. Unlike
+records, they describe things that can't or shouldn't be copied by value.
+Instead their ownership can be passed between two components via a
+[handle](#handles). Resources are similar to a a `class` in many programming
+languages and can have constructors and methods.
+
+For example, the following Wit defines a resource type and a function that takes
+and returns a handle to a `blob`:
+```wit
+resource blob;
+transform: func(blob) -> blob;
+```
+
+As syntactic sugar, resource statements can also declare any number of
+*methods*, which are functions that implicitly take a `self` parameter that is a
+handle. A resource statement can also contain any number of *static functions*,
+which do not have an implicit `self` parameter but are meant to be lexically
+nested in the scope of the resource type. Lastly, a resource statement can
+contain at most one *constructor* function, which is syntactic sugar for a
+function returning a handle of the containing resource type.
+
+For example, the following is a resource with one constructor, two methods, and
+a static function:
+
+```wit
+resource blob {
+    constructor(init: list<u8>);
+    write: func(bytes: list<u8>);
+    read: func(n: u32) -> list<u8>;
+    merge: static func(lhs: borrow<blob>, rhs: borrow<blob>) -> blob;
+}
+```
+
+Methods always desugar to a borrowed self parameter whereas constructors always
+desugar to an owned return value. For example, the `blob` constructor
+[above](#resources) desugars to:
+
+```wit
+resource blob;
+%[constructor]blob: func(self: borrow<blob>, bytes: list<u8>) -> blob;
+%[method]blob.write: func(self: borrow<blob>, bytes: list<u8>);
+%[method]blob.read: func(self: borrow<blob>, n: u32) -> list<u8>;
+%[static]blob.merge: func(lhs: borrow<blob>, rhs: borrow<blob>) -> blob;
+```
+
+#### Handles
+
+There are two types of handles in Wit: "owned" handles and "borrowed" handles.
+Owned handles represent the passing of unique ownership of a resource between
+two components. When the owner of an owned handle drops that handle, the
+resource is destroyed. In contrast, a borrowed handle represents a temporary
+loan of a handle from the caller to the callee for the duration of the call.
+
+When a resource type name is used directly (e.g. when `blob` is used as the
+return value of the `blob` constructor [above](#resources)), it stands for an
+"owning" handle that will call the resource's destructor when dropped. When a
+resource type name is wrapped with `borrow<...>`, it stands for a "borrowed"
+handle that will *not* call the destructor when dropped.
+
 ### Flags
 
-A flags type is a set of named booleans.  In an instance of the type, each flag will be either true or false.
+A flag type is a set of named booleans.  In an instance of the type, each flag will be either true or false.
 
 ```wit
 flags allowed-methods {
--- a/component-model/src/design/wit.md
+++ b/component-model/src/design/wit.md
@@ -209,7 +209,8 @@ Resources are handles to some entity that lives outside of the component. They
 describe things that can't or shouldn't be copied by value; instead, their
 ownership or reference can be passed between two components via a handle. Unlike
 other WIT types which are simply plain data, resources only expose behavior
-through methods. Resources are _objects that implement_ an interface.
+through methods. Resources can be thought of as _objects that implement_ an
+interface.
 
 For example, we could model a blob (binary large object) as a resource. The
 following WIT defines the a `blob` resource type, which contains a constructor,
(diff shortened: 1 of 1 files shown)


component-model/src/design/wit.md
```diff
@@ -202,9 +204,70 @@ enum color {
...
+
+For example, the following is a resource with one constructor, two methods, and
+a static function:
+
+```wit
+resource blob {
+    constructor(init: list<u8>);
+    write: func(bytes: list<u8>);
+    read: func(n: u32) -> list<u8>;
+    merge: static func(lhs: borrow<blob>, rhs: borrow<blob>) -> blob;
```
itowlson, line 238: Maybe add a comment reassuring the reader not to worry about `borrow` for now and it will be explained below?

component-model/src/design/wit.md
```diff
@@ -202,9 +204,70 @@ enum color {
...
+}
+```
+
+Methods always desugar to a borrowed self parameter whereas constructors always
```
itowlson, line 242: I found this paragraph hard to parse - it felt like I'd suddenly been dropped into a Rust compiler implementors' chat.  What is borrowing?  Do I need to care about desugaring?  What do you want me as a _user_ of resources, as opposed to someone implementing them in a Wasm runtime, to understand here?

component-model/src/design/wit.md
```diff
@@ -202,9 +204,70 @@ enum color {
...
+desugar to an owned return value. For example, the `blob` constructor
+[above](#resources) desugars to:
```
itowlson, line 244: Might want some more subheadings or anchors if we want to do links _within_ this section?

component-model/src/design/wit.md
```diff
@@ -202,9 +204,70 @@ enum color {
...
+
+```wit
+resource blob;
+%[constructor]blob: func(self: borrow<blob>, bytes: list<u8>) -> blob;
```
itowlson, line 248: Under what circumstances do application developers need to know about the mangled forms?  If we have a "how resources work" section then this would be great to have there but I'd keep it out of the WIT guide.

component-model/src/design/wit.md
```diff
@@ -202,9 +204,70 @@ enum color {
...
+When a resource type name is used directly (e.g. when `blob` is used as the
+return value of the `blob` constructor [above](#resources)), it stands for an
+"owning" handle that will call the resource's destructor when dropped. When a
+resource type name is wrapped with `borrow<...>`, it stands for a "borrowed"
+handle that will *not* call the destructor when dropped.
+
 ### Flags
 
-A flags type is a set of named booleans.  In an instance of the type, each flag will be either true or false.
+A flag type is a set of named booleans.  In an instance of the type, each flag will be either true or false.
```
itowlson, line 270: `flags` here was intentional because the type expresses a _set_ of flags not a single flag.

component-model/src/design/wit.md
```diff
@@ -202,9 +204,70 @@ enum color {
...
+resource blob;
+%[constructor]blob: func(self: borrow<blob>, bytes: list<u8>) -> blob;
+%[method]blob.write: func(self: borrow<blob>, bytes: list<u8>);
+%[method]blob.read: func(self: borrow<blob>, n: u32) -> list<u8>;
+%[static]blob.merge: func(lhs: borrow<blob>, rhs: borrow<blob>) -> blob;
+```
+
+#### Handles
+
+There are two types of handles in Wit: "owned" handles and "borrowed" handles.
```
itowlson, line 256:
```suggestion
There are two types of handles in WIT: "owned" handles and "borrowed" handles.
```

itowlson, line 256: Oh, okay, so resources _only ever_ surface as handles!  I think it would be good to explain that closer to where resources and handles are introduced, maybe with some suitable analogy.

(also the earlier text said handles represent ownership so having this additional "may represent ownership or borrowed reference" up there would help set the reader on the right track)

component-model/src/design/wit.md
```diff
@@ -202,9 +204,70 @@ enum color {
...
+Owned handles represent the passing of unique ownership of a resource between
+two components. When the owner of an owned handle drops that handle, the
+resource is destroyed. In contrast, a borrowed handle represents a temporary
+loan of a handle from the caller to the callee for the duration of the call.
+
+When a resource type name is used directly (e.g. when `blob` is used as the
+return value of the `blob` constructor [above](#resources)), it stands for an
+"owning" handle that will call the resource's destructor when dropped. When a
+resource type name is wrapped with `borrow<...>`, it stands for a "borrowed"
+handle that will *not* call the destructor when dropped.
```
itowlson, line 266: Oof, I can see a lot of readers finding this tricky - not because of what you've written, but because the design is so predicated on Rust-like notions of ownership and borrowing.  "So is a borrow like a pointer?  Yes, but owning is like a pointer too..."  I don't have any good ideas for explaining it, though!  But let's get this in and we can keep iterating on it!

component-model/src/design/wit.md
```diff
@@ -202,9 +204,70 @@ enum color {
...
+
+There are two types of handles in Wit: "owned" handles and "borrowed" handles.
+Owned handles represent the passing of unique ownership of a resource between
+two components. When the owner of an owned handle drops that handle, the
+resource is destroyed. In contrast, a borrowed handle represents a temporary
+loan of a handle from the caller to the callee for the duration of the call.
+
+When a resource type name is used directly (e.g. when `blob` is used as the
+return value of the `blob` constructor [above](#resources)), it stands for an
+"owning" handle that will call the resource's destructor when dropped. When a
```
itowlson, line 264: This somewhat recapitulates the distinction described in the previous paragraph.  Maybe merge the two paragraphs, possibly as a bullet of "owned means ... and has X characteristics / borrowed means ... and has Y characteristics"?

component-model/src/design/wit.md
```diff
@@ -202,9 +204,70 @@ enum color {
...
+%[method]blob.write: func(self: borrow<blob>, bytes: list<u8>);
+%[method]blob.read: func(self: borrow<blob>, n: u32) -> list<u8>;
+%[static]blob.merge: func(lhs: borrow<blob>, rhs: borrow<blob>) -> blob;
+```
+
+#### Handles
+
+There are two types of handles in Wit: "owned" handles and "borrowed" handles.
+Owned handles represent the passing of unique ownership of a resource between
+two components. When the owner of an owned handle drops that handle, the
```
itowlson, line 258: `drop` is Rustspeak.  Can we translate / analogise it?  Should we break out a section on resource lifetimes / lifecycles?

(or maybe that is a matter for future iteration)

component-model/src/design/wit.md
```diff
@@ -202,9 +204,70 @@ enum color {
 
 This can provide a simpler representation in languages without discriminated unions. For example, a WIT enum can translate directly to a C++ `enum`.
 
+### Resources
+
+A resource type defines a new abstract type an entity with a lifetime. Unlike
```
itowlson, line 209: Another important characteristic of an entity/resource is identity.  If I hand over a `Point` record to a callee, the copy in the callee is a different record and can be manipulated separately.  If I hand over a `Car` resource, the callee is using _my_ `Car` - there's only one actual `Car`.  Like if the borrower drives its reference into a lamp post, my windscreen gets broken.

itowlson (changes requested):
This is great - thank you so much for wading into a difficult and nuanced topic.  I'm afraid I had a few areas where I got confused, some of which is down to the design and we probably need to punt on those for this PR and iterate on them over time, but some of which can hopefully addressed by some reorganising.  But this is an area where every concept seems to depend on every other concept, so I may be being optimistic... if it gets problematic then I'm totally up for landing this as is and continuing to noodle on it as the community learns what sort of explanations work best.

component-model/src/design/wit.md
kate-goldenring: I almost did add a section on handles. Should i simplify all of this and remove any deep dive into handles?

component-model/src/design/wit.md
kate-goldenring: Most of this I directly pulled over from WIT.md, which is likely a more technical audience. We could pull some of this out

component-model/src/design/wit.md
itowlson: It seems like there needs to be something about borrowed vs owned, but yeah, it's worth seeing if just calling things "resources" would work.  (Although it might be good to say "the spec distinguishes between resources and handles, but that's a runtime implementation concern," in case someone reads the spec and is wondering why it talks about things differently.)

Re the broader question, to me this is meant to be a "what you need to know to write WITs for your components (or read WITs for somebody else's components" (and I haven't looked at it since the initial writing frenzy so it may not be well aligned to that vision).  That reflects the idea of the book as being for readers trying to _get stuff done_ with the component model, rather than a re-layout or curation of the design docs.  But that's definitely a question that should be addressed by the wider SIG - I'm only one voice for sure!

kate-goldenring: I think I am aligned with the idea of provide enough information to get stuff done. And this doc is about syntactical keywords, which `handle` is not one of. I removed most references to `handle` and simplified the section. Do you mind reviewing it again?

component-model/src/design/wit.md
```diff
@@ -17,7 +17,7 @@ The WIT (Wasm Interface Type) language is used to define Component Model interfa
     - [Records](#records)
     - [Variants](#variants)
     - [Enums](#enums)
-    - [Flags](#flags)
+    - [Resources](#resources)
```
itowlson, line 20: Hiccup: Flags has disappeared from TOC

component-model/src/design/wit.md
itowlson: I like it!  I don't think anything significant to app developers is lost in the simplification (though I may turn out to be wrong), and I definitely felt more comfortable with the concepts this time around.  Thank you!

component-model/src/design/wit.md
kate-goldenring: Good point. I updated this to say "it could also be written as" and expanded the function signatures. I will say i am not positive this expansion of methods is syntactically correct. @lann @dicej, are these equivalent?:

```wit
resource blob {
    constructor(init: list<u8>);
    write: func(bytes: list<u8>);
    read: func(n: u32) -> list<u8>;
    merge: static func(lhs: borrow<blob>, rhs: borrow<blob>) -> blob;
}
 // Verses
resource blob {
    blob: func(self: borrow<blob>, bytes: list<u8>) -> blob;
    write: func(self: borrow<blob>, bytes: list<u8>);
    read: func(self: borrow<blob>, n: u32) -> list<u8>;
    merge: static func(lhs: borrow<blob>, rhs: borrow<blob>) -> blob;
}
```

I am trying to illustrate how it is desugared to the following without showing that level of complexity:
```wit
resource blob;
%[constructor]blob: func(self: borrow<blob>, bytes: list<u8>) -> blob;
%[method]blob.write: func(self: borrow<blob>, bytes: list<u8>);
%[method]blob.read: func(self: borrow<blob>, n: u32) -> list<u8>;
%[static]blob.merge: func(lhs: borrow<blob>, rhs: borrow<blob>) -> blob;
```

component-model/src/design/wit.md
kate-goldenring: Thanks! Fixed

component-model/src/design/wit.md
```diff
@@ -202,6 +203,54 @@ enum color {
 
 This can provide a simpler representation in languages without discriminated unions. For example, a WIT enum can translate directly to a C++ `enum`.
 
+### Resources
+
+A resource type defines a new abstract type: an entity with ownership. Unlike
+records, they describe things that can't or shouldn't be copied by value.
+Instead their ownership or reference can be passed between two components via a
+handle. Resources are similar to a a `class` in many programming
+languages and can have constructors and methods.
```
lann, line 212: Might want to add a callout here with a little more information, maybe something along the lines of:

> Unlike all other WIT types, a resource type is unique to the instance that defines it. Two resource types with the same name defined in different components - even two instances of the same component - are distinct types.

component-model/src/design/wit.md
```diff
@@ -202,6 +203,54 @@ enum color {
...
+
+- _methods_: functions that implicitly take a `self` (similar to `this`)
+parameter that is a handle
+- _static functions_: functions which do not have an implicit `self` parameter
+but are meant to be lexically nested in the scope of the resource type
+- at most one _constructor_: a function that is syntactic sugar for a function
+returning a handle of the containing resource type
+
+Methods always desugar to a borrowed `self` parameter whereas constructors
+always desugar to an owned return value. For example, the `blob` constructor
```
lann, line 237:
```suggestion
always desugar to an owned return value. For example, the `blob` resource
```

component-model/src/design/wit.md
itowlson: Is there a way to characterise that in terms of component developer implications because I opened that link and now my brain is hiding under the sofa

component-model/src/design/wit.md
```diff
@@ -202,6 +203,54 @@ enum color {
...
+[above](#resources) could also be written as:
```
lann, line 238: I think this could be interpreted as saying that the two definitions are _identical_ which would be misleading.

Hard to get the language exactly right here without being too wordy or formal, but maybe something like:

```suggestion
[above](#resources) could be approximated as:
```

component-model/src/design/wit.md
```diff
@@ -202,6 +203,54 @@ enum color {
...
+
+```wit
+resource blob {
+    blob: func(self: borrow<blob>, bytes: list<u8>) -> blob;
+    write: func(self: borrow<blob>, bytes: list<u8>);
+    read: func(self: borrow<blob>, n: u32) -> list<u8>;
+    merge: static func(lhs: borrow<blob>, rhs: borrow<blob>) -> blob;
+}
```
lann, line 246: I know this has seen a few iterations already so feel free to ignore; it might still be worth splitting these out into separate functions _without_ getting into name mangling:

```suggestion
resource blob;
blob: func(self: borrow<blob>, bytes: list<u8>) -> blob;
write: func(self: borrow<blob>, bytes: list<u8>);
read: func(self: borrow<blob>, n: u32) -> list<u8>;
merge: static func(lhs: borrow<blob>, rhs: borrow<blob>) -> blob;

```

component-model/src/design/wit.md
```diff
@@ -202,6 +203,54 @@ enum color {
...
+- at most one _constructor_: a function that is syntactic sugar for a function
+returning a handle of the containing resource type
+
+Methods always desugar to a borrowed `self` parameter whereas constructors
+always desugar to an owned return value. For example, the `blob` constructor
+[above](#resources) could also be written as:
+
+```wit
+resource blob {
+    blob: func(self: borrow<blob>, bytes: list<u8>) -> blob;
```
lann, line 242:
```suggestion
    blob: func(bytes: list<u8>) -> blob;
```

lann: Separately, I would maybe call this `blob-constructor` just so it doesn't seem like constructors are literally named after the resource :shrug:

component-model/src/design/wit.md
lann: I think I'd just like to balance "these are kind of like classes" with a little "but only kind of; beware".

itowlson: Yeah but my questions about "can I pass them, where do I define them to do so" were not rhetorical - given what you're saying, I think we genuinely need to clarify that to developers.  Or am I missing something?  I am clearly missing something. _grin_

dicej: I tend to think of resources as interfaces (or traits or protocols or abstract base classes or whatever term you're used to) rather than classes: there can (and usually will) be multiple implementations, and those implementations are not interchangeable.  You can certainly pass handles to a given implementation within and across components, though.

lann: Yes you can pass them around. in order to work with (or even receive) a handle to a particular resource a component has to import that exact resource type, either directly from the component instance that defined it or indirectly via aliases/re-exports.

I think it ends up being fairly intuitive most of the time though with the way interfaces are reused in wit it could be confusing that two exports of the same interface produce incompatible resource handles.

lann: Yes by default all imports will end up with the same instance if you aren't explicitly doing something "weird".

component-model/src/design/wit.md
kate-goldenring: @lann, the WIT explainer has the constructor also taking in a reference to self:
`%[constructor]blob: func(self: borrow<blob>, bytes: list<u8>) -> blob;`. Is that an error?

lann: Yeah that's an error. https://github.com/WebAssembly/component-model/pull/322

component-model/src/design/wit.md
```diff
@@ -1,6 +1,6 @@
 # An Overview of WIT
 
-The WIT (Wasm Interface Type) language is used to define Component Model interfaces and worlds. WIT isn't a general-purpose coding language and doesn't define behaviour; it defines only _contracts_ between components. This topic provides an overview of key elements of the WIT language.
+The WIT (Wasm Interface Type) language is used to define Component Model interfaces and worlds. WIT isn't a general-purpose coding language and doesn't define behaviour; it defines only _contracts_ between components. This topic provides an overview of key elements of the WIT language. The official WIT specification and history can be found in the [`WebAssembly/component-model` repository](https://github.com/WebAssembly/component-model/blob/main/design/mvp/WIT.md).
```
rylev, line 3: > The WIT (Wasm Interface Type) language is used to define Component Model interfaces and worlds.

I know this was already here (so feel free to ignore), but do we expect people to know what "worlds" are? I wonder if this can be phrased in a way that doesn't yet use the specific vernacular of the component model but still indicates the overall shape of what we're talking about.

component-model/src/design/wit.md
```diff
@@ -202,6 +203,54 @@ enum color {
 
 This can provide a simpler representation in languages without discriminated unions. For example, a WIT enum can translate directly to a C++ `enum`.
 
+### Resources
+
+A resource type defines a new abstract type: an entity with ownership. Unlike
+records, they describe things that can't or shouldn't be copied by value.
```
rylev, line 209: This feels a bit jargony. I'm not quite sure what the right wording would be, but perhaps something like:

> Resources are handles to some entity that lives outside of the component. Unlike other WIT types which are simply plain data, resources only expose behavior through methods.

component-model/src/design/wit.md
```diff
@@ -202,6 +203,54 @@ enum color {
...
+Instead their ownership or reference can be passed between two components via a
+handle. Resources are similar to Rust `trait`s or C# and TypeScript `interface`s
```
rylev, line 211: I don't think it's quite right to say resources are like Rust's traits. They are exactly like trait objects (e.g., `Box<dyn MyTrait>`). Where are traits are description of some interface, trait objects are an opaque reference to an implementation of that interface - the same as resources.

In TypeScript and C#, I think resources are also not quite interfaces. They are objects which implement an interface.

component-model/src/design/wit.md
```diff
@@ -202,6 +203,54 @@ enum color {
...
+    constructor(init: list<u8>);
+    write: func(bytes: list<u8>);
+    read: func(n: u32) -> list<u8>;
+    merge: static func(lhs: borrow<blob>, rhs: borrow<blob>) -> blob;
+}
+```
+
+As shown in the `blob` example, a resource can contain:
+
+- _methods_: functions that implicitly take a `self` (similar to `this`)
```
rylev, line 230: > (similar to `this`)

I think we can just say:

> (often called `this` in many languages)

component-model/src/design/wit.md
rylev: I would personally not mention `borrow` here. It risks confusing the reader. Can we just have this method take ownership of the `lhs` and `rhs`?

component-model/src/design/wit.md
```diff
@@ -202,6 +203,54 @@ enum color {
...
+parameter that is a handle
+- _static functions_: functions which do not have an implicit `self` parameter
+but are meant to be lexically nested in the scope of the resource type
```
rylev, line 233:
```suggestion
but are meant to be nested in the scope of the resource type
```

I'm not sure `lexically` adds too much and can be jargony and confusing.

component-model/src/design/wit.md
```diff
@@ -202,6 +203,54 @@ enum color {
...
+
+Methods always desugar to a borrowed `self` parameter whereas constructors
+always desugar to an owned return value. For example, the `blob` resource
+[above](#resources) could be approximated as:
+
+```wit
+resource blob;  
+blob-constructor: func(bytes: list<u8>) -> blob;  
+blob-write: func(self: borrow<blob>, bytes: list<u8>);  
+blob-read: func(self: borrow<blob>, n: u32) -> list<u8>;
```
rylev, line 245: Has `borrow<T>` been explained at all?

component-model/src/design/wit.md
```diff
@@ -202,6 +203,54 @@ enum color {
...
+As shown in the `blob` example, a resource can contain:
+
+- _methods_: functions that implicitly take a `self` (similar to `this`)
+parameter that is a handle
+- _static functions_: functions which do not have an implicit `self` parameter
+but are meant to be lexically nested in the scope of the resource type
+- at most one _constructor_: a function that is syntactic sugar for a function
+returning a handle of the containing resource type
+
+Methods always desugar to a borrowed `self` parameter whereas constructors
```
rylev, line 237: I think we need to be more verbose in our explanation of this. While this explanation probably makes a lot of sense to Rust and C programmers, it probably doesn't make any sense to programmers of GCed languages.

component-model/src/design/wit.md
kate-goldenring: But then what is a `handle`? I am trying to avoid too many concepts, but i get that we can use the word handle without refering to a `handle` in the component model sense. I did originally have a section on `handle`s but we decided to remove a section in this thread https://github.com/bytecodealliance/component-docs/pull/103#discussion_r1523937953 to simplify.

component-model/src/design/wit.md
kate-goldenring: There is a discussion on what analogy to draw in this thread: https://github.com/bytecodealliance/component-docs/pull/103#discussion_r1523973325. I will try to rewrite this with this feedback in mind, but i am coming to the conclustion that no analogy can be drawn

kate-goldenring: I removed any analogy and worked up to emphasizing they are "objects that implement" interfaces.

component-model/src/design/wit.md
kate-goldenring: We had a note about it, but i moved it out of the note into a paragraph below this text. Let me know what you think

component-model/src/design/wit.md
kate-goldenring: Can we leave this to a future iteration? There is a lot of feedback i am juggling here and i fear that digging into this deeper will create another round of feedback that may be better explored in it's own PR. I would like to get this in before KubeCon WasmDay.

component-model/src/design/wit.md
kate-goldenring: In the least, this is a good call out to link to the docs on `worlds`. I'd rather link to more documentation on terms than avoid terms.

component-model/src/design/wit.md
kate-goldenring: I'd love to punt this to a future iteration. Happy to capture this feedback into an issue we can track on explaining the terms more generically

kate-goldenring:
@rylev @itowlson do you mind giving this another look? There is a lot of great feedback here. I'd love to get something in for starters and iterate from there.

itowlson (approved):
LGTM - I think this is a great distillation of the essentials, and as you say, we can iterate on the details over time.  Thanks for your patience balancing the different viewpoints here.

Issue #84, opened 2023-11-06
Python tooling docs fail with NotImplmentedError


sachitv:
I was following the instructions here: https://component-model.bytecodealliance.org/language-support/python.html#running-components-from-python-applications

I have tried using wasmtime from the fork mentioned there (dicej/wasmtime-py) as well as wasmtime 14.0.0 but I encounter failures when running wasmtime.bindgen.
**Environment**
Ubuntu 22.04 (amd64) also reproduced on Mac OS Sonoma (14.1) (arm64)
Poetry 1.7.0
Python 3.12.0 via pyenv

**Command:**
`python3 -m wasmtime.bindgen add.wasm --out-dir add`

**Failure:**
```
Traceback (most recent call last):
  File "/home/sachit/Development/learn-componentize/.venv/lib/python3.12/site-packages/wasmtime/_func.py", line 260, in enter_wasm
    yield byref(trap)
  File "/home/sachit/Development/learn-componentize/.venv/lib/python3.12/site-packages/wasmtime/_func.py", line 101, in __call__
    raise WasmtimeError._from_ptr(error)
wasmtime._error.WasmtimeError: error while executing at wasm backtrace:
    0:  0x100 - wit-component:shim!indirect-wasi:cli/terminal-stdin-get-terminal-stdin
    1: 0x254f - wit-component:adapter:wasi_snapshot_preview1!wasi_snapshot_preview1::descriptors::Descriptors::new::ha2fdde51d30a71bd
    2: 0x1640 - wit-component:adapter:wasi_snapshot_preview1!wasi_snapshot_preview1::State::descriptors::h780c546b61bcfc7f
    3: 0x180c - wit-component:adapter:wasi_snapshot_preview1!fd_write
    4:  0x13a - wit-component:shim!adapt-wasi_snapshot_preview1-fd_write
    5: 0x1b080d - <unknown>!wasi::lib_generated::fd_write::hd4964fea612b930f
    6: 0x1acdfa - <unknown>!<std::io::Write::write_fmt::Adapter<T> as core::fmt::Write>::write_str::h375f1d6863bea9df
    7: 0x1ba82b - <unknown>!core::fmt::write::h0eddb54b80b97b9d
    8: 0x1add1c - <unknown>!std::io::Write::write_fmt::h6d46415105134b08
    9: 0x1afaa2 - <unknown>!std::panicking::default_hook::{{closure}}::he04c18047097e21e
[...]

PR #106, opened 2024-03-15
fix: bring documentation and tutorials up to date with latest tooling

kate-goldenring:
A lot of the language guides were out of date. This updates:
- The example host tutorial to use latest wasmtime
- removes unnecessary component model flag for `wasmtime`
- removes two files that were stranded a while ago (under `/implementations` ) when they were moved to `runtimes`
- calls out `componentize-py` and `wasmtime-py` compatibility to avoid this issue https://github.com/bytecodealliance/component-docs/issues/84

I ran through everything and got it working. Ideally we can get this in and iterate if there is a lot of syntactical feedback


component-model/src/language-support/javascript.md
```diff
@@ -3,17 +3,26 @@
 [`jco`](https://github.com/bytecodealliance/jco) is a fully native JS tool for working with the
 emerging WebAssembly Components specification in JavaScript.
 
-### Building a Component with `jco`
+## Building a Component with `jco`
 
-A component can be created from a JS module using `jco componentize`. First, install `jco` and
-`componentize-js`:
+A component can be created from a JS module using `jco componentize`. First, install `jco`:
```
itowlson, line 8: This paragraph removes the instruction to install `componentize-js`, but that instruction is still shown in the `npm` commands just below.  Which is correct?

component-model/src/language-support/python.md
```diff
@@ -36,41 +57,100 @@ $ cargo run --release -- 1 2 ../path/to/add.wasm
 1 + 2 = 3
 ```
 
-### Running components from Python Applications
+See [`componentize-py`'s examples](https://github.com/bytecodealliance/componentize-py/tree/main/examples) to try out build HTTP, CLI, and TCP components from Python applications.
+
+
+### Building a Component that exports an interface
```
itowlson, line 63: Nit: capitalisation in this title is a bit random (as, admittedly, it was in the previous iteration).

(Longer term we should pick either Sentence case or Title Case and align all our headings around it.  But that's definitely out of scope for this PR!)

component-model/src/language-support/python.md
```diff
@@ -36,41 +57,100 @@ $ cargo run --release -- 1 2 ../path/to/add.wasm
...
+
+The [sample `add.wit` file](https://github.com/bytecodealliance/component-docs/tree/main/component-model/examples/example-host/add.wit) exports a function. However, to use your component from another component, it must export an interface. That being said, you rarely find WIT that does not contain an interface. Let's expand our example world to export an interface rather than directly export the function.
```
itowlson, line 65:
```suggestion
The [sample `add.wit` file](https://github.com/bytecodealliance/component-docs/tree/main/component-model/examples/example-host/add.wit) exports a function. However, to use your component from another component, it must export an interface. (Most WITs you'll see in the wild do use interfaces; we've been simplifying by exporting a function.)  Let's expand our example world to export an interface rather than directly export the function.
```

Or something like that.  Just trying to address the "wait why has the book been showing us the _exact thing_ it now says we'll rarely find" reaction.

component-model/src/language-support/python.md
```diff
@@ -36,41 +57,100 @@ $ cargo run --release -- 1 2 ../path/to/add.wasm
...
+interface add {
+    add: func(a: u32, b: u32) -> u32;
+}
+
+world example {
+    export add;
+}
+```
+
+If you peak at the bindings, you'll notice that we now implement a class for the `add` interface rather than for the `example` world. This is a consistent pattern. As you export more interfaces from your world, you implement more classes. Our add example gets the slight update of:
```
itowlson, line 80:
```suggestion
If you peek at the bindings, you'll notice that we now implement a class for the `add` interface rather than for the `example` world. This is a consistent pattern. As you export more interfaces from your world, you implement more classes. Our add example gets the slight update of:
```

pet peeve alert

component-model/src/language-support/python.md
```diff
@@ -36,41 +57,100 @@ $ cargo run --release -- 1 2 ../path/to/add.wasm
...
+
+```py
+# app.py
+import example
+
+class Add(example.Example):
```
itowlson, line 86: The explanatory paragraph said "we now implement a class for the `add` interface rarher than the `example` world, but this still inherits from `example.Example`.  It would be good to clarify what makes this implement that specific interface (of all the ones in the `example` world).  Is it the class name?

component-model/src/language-support/python.md
```diff
@@ -36,41 +57,100 @@ $ cargo run --release -- 1 2 ../path/to/add.wasm
...
 ```
 
-Now, generate the bindings to be able to call the component from a Python host application.
+First, generate the bindings to be able to call the component from a Python host application.
 
 ```sh
+# Get an add component that does not import the WASI CLI world
+$ wget https://github.com/bytecodealliance/component-docs/raw/main/component-model/examples/example-host/add.wasm
+# Validate the component against the checksum
+$ echo "7a64aa30ac7d91b03448001dcdc23760a80c58ecaf220bc70998d49b88820440  add.wasm" | sha256sum --check
```
itowlson, line 117: This feels like it could easy slip out of sync if we ever rebuild `add.wasm`.

component-model/src/language-support/python.md
```diff
@@ -36,41 +57,100 @@ $ cargo run --release -- 1 2 ../path/to/add.wasm
...
+    
+    def __init__(self, store: wasmtime.Store) -> None:
+        path = pathlib.Path(__file__).parent / ('root.core0.wasm')
+        module = wasmtime.Module.from_file(store.engine, path)
+        instance0 = wasmtime.Instance(store, module, []).exports(store)
+        lift_callee0 = instance0["add"]
+        assert(isinstance(lift_callee0, wasmtime.Func))
+        self.lift_callee0 = lift_callee0
+    def add(self, caller: wasmtime.Store, x: int, y: int) -> int:
+        ret = self.lift_callee0(caller, _clamp(x, -2147483648, 2147483647), _clamp(y, -2147483648, 2147483647))
```
itowlson, line 140: Yikes!  Does the user need to understand all this stuff?!  Could we elide it, or move it to an "internals" section?  If not, can we clarify to readers what they should be looking at?

itowlson (changes requested):
This is great - it is so good to finally see the tooling starting to smooth out its rough edges.  And I love how you've made the flow through the different tutorials more consistent.

My comments are mostly nits except for the one on the printout of the generated Python bindings.  I bounced off that pretty hard I'm afraid.  But the rest is great.

component-model/src/language-support/python.md
kate-goldenring: Yeah maybe we add a style guide

component-model/src/language-support/python.md
kate-goldenring: Good point. I also have an issue up to switch to using an interface in a different iteration

component-model/src/language-support/python.md
kate-goldenring: 🙈

component-model/src/language-support/python.md
kate-goldenring: Should we remove the checksum since it is a module rather than an executable binary? Or we can add a `add-checksum` file next to it to make it more obvious

component-model/src/language-support/python.md
kate-goldenring: This is all generated code that i wanted to show them in the docs. Normally in the docs, we have not shown the bindings that are generated -- so I will remove this section

component-model/src/language-support/python.md
itowlson: I'd probably remove the checksum because I have a childlike faith in the Wasm sandbox.  Having a checksum file available for people who want it is probably a decent plan, as long as we can automate maintenance.  Is the .wasm file built in CI or do authors build locally and check it in?

kate-goldenring: > authors build locally and check it in
this one. Can can have CI for it in the future.

kate-goldenring: I removed the checksum

PR #109, opened 2024-03-18
Improve the 'worlds' explainer

rylev:
An attempt to improve the worlds explainer. Happy to justify any change if it's not clear why. Some examples of changes include:
* Using the term "target" for what application/library developers due and explicitly contrasting that with "supporting" for environments to that run components.
* Explicitly mention that a component can target a world that either has been defined by the component author or someone else.
* Use the wasi command line example to illustrate both a runtime that supports that world *and* authoring a component that targets that world
* Remove the redundant wasi command line world example


component-model/src/design/worlds.md
```diff
@@ -2,16 +2,19 @@
...
+On one hand, a world describes the shape of a component - it says which interfaces the component exposes for other code to call (its exports), and which interfaces the component depends on (its imports). A world only defines the surface of a component, not the internal behaviour. If you're an application or library developer creating a component, you'll specify the world your component targets which describes the functionality your component exposes to the outside world and declares the functionality your component depends on in order to be able to run. Your component many target a custom world definition you have created with a unique set of imports and exports tailored just for your use case, or it may target an existing world definition that someone else has already specified.
```
kate-goldenring, line 5: "to the outside world" may be confusing or a good pun?
```suggestion
On one hand, a world describes the shape of a component - it says which interfaces the component exposes for other code to call (its exports), and which interfaces the component depends on (its imports). A world only defines the surface of a component, not the internal behaviour. If you're an application or library developer creating a component, you'll specify the world your component targets. This world describes the functionality your component exposes and declares the functionality your component depends on in order to be able to run. Your component many target a custom world definition you have created with a unique set of imports and exports tailored just for your use case, or it may target an existing world definition that someone else has already specified.
```

component-model/src/design/worlds.md
```diff
@@ -2,16 +2,19 @@
...
 
-For a component to run, its imports must be fulfilled, by a host or by other components.  Connecting up one component's imports to another component's matching exports is called _composition_.
+A world is composed of interfaces, but each interface is _directional_ - it indicates whether the interface is available for outside code to call (an "export"), or outside code must fulfill the interface for the component to call (an "import"). These interfaces strictly bound the component. A component cannot interact with anything outside itself except by having its exports called, or by it calling its imports. This provides very strong sandboxing; for example, if a component does not have an import for a secret store, then it _cannot_ access that secret store, even if the store is running in the same process.
```
kate-goldenring, line 11:
```suggestion
A world is composed of interfaces, but each interface is _directional_ - it indicates whether the interface is available for outside code to call (an "export"), or whether outside code must fulfill the interface for the component to call (an "import"). These interfaces strictly bound the component. A component cannot interact with anything outside itself except by having its exports called, or by it calling its imports. This provides very strong sandboxing; for example, if a component does not have an import for a secret store, then it _cannot_ access that secret store, even if the store is running in the same process.
```

component-model/src/design/worlds.md
```diff
@@ -2,16 +2,19 @@
...
 
-* A (trivial) "HTTP proxy" world would export a "receive HTTP requests" interface, and import a "send HTTP requests" interface. A host, or another component, would call the exported "receive" interface, passing an HTTP request; the component would forward it on via the imported "send" interface. To be a _useful_ proxy, the component may also need to import interfaces such as I/O and clock time - without those imports the component could not perform, for example, on-disk caching.
-* The "WASI (WebAssembly System Interface) command line" world is a classic example of a "hosting environment" use of the world concept. This world exports APIs such as file I/O, sockets, random number generation, and other POSIX-style functionality, so that application components that depend on this world - that is, command line applications - can use those familiar capabilities.
+For a component to run, its imports must be fulfilled, by a host or by other components. Connecting up one component's imports to another component's matching exports is called _composition_.
```
kate-goldenring, line 13: Should we note somewhere that the imports can be satisfied by composition with N many components -- that it does not have to be 1 + 1 = fulfilled component.

kate-goldenring (commented):
@rylev I really like how you break down the command line world. @itowlson wrote the original version, so I'll leave it to him to approve but LGTM.

itowlson (approved):

PR #110, opened 2024-03-18
Improve Rust language guide

rylev:
Small changes to the Rust language guide. Let me know if any particular change doesn't make sense, and I will provide a justification. Some example changes include:
* Note that the bindings are a generated module
* Explain in more detail why the `calculator.wasm` example cannot run its non-composed form.

### DIFF
--- a/component-model/src/language-support/rust.md
+++ b/component-model/src/language-support/rust.md
@@ -16,14 +16,15 @@ cargo install cargo-component
 ## Building a Component with `cargo component`
 
 Create a Rust library that implements the `add` function in the [`example`
-world](https://github.com/bytecodealliance/component-docs/tree/main/component-model/examples/example-host/add.wit). Note that it imports the bindings that will be created by
-`cargo-component`. First scaffold a project:
+world](https://github.com/bytecodealliance/component-docs/tree/main/component-model/examples/example-host/add.wit). First scaffold a project:
 
 ```sh
 $ cargo component new add --lib && cd add
 ```
 
-Update `wit/world.wit` to match `add.wit` and modify the component package reference to change the
+Note that `cargo component` generates the necessary bindings as a module called `bindings`. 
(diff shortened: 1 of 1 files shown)


component-model/src/language-support/rust.md
```diff
@@ -191,7 +190,7 @@ world root {
 }
 ```
 
-As the import is unfulfilled, the `calculator.wasm` component could not run by itself in its current form. To fulfil the `add` import, so that the calculator can run, you would need to [compose the `calculator.wasm` and `adder.wasm` files into a single, self-contained component](../creating-and-consuming/composing.md).
+As the world that the [`example-host` targets](https://github.com/bytecodealliance/component-docs/tree/main/component-model/examples/example-host/add.wit) does not include the `docs:adder/add@0.1.0` import, the example host cannot run the `calculator.wasm` component in its current form - `calculator.wasm` expects to be able to call `add` but the example host does not provide that functionality. To fulfil the `add` import, so that the calculator can run, you can [compose the `calculator.wasm` and `adder.wasm` files into a single, self-contained component](../creating-and-consuming/composing.md).
```
kate-goldenring, line 193: The example host never executes the calculator. That host is only for the `add` world. The tutorial discussed composing the calculator world with a command world in order to execute it but we do not provide any hosts in the docs for the `calculator` world (that call `calculate`).
```suggestion
As the import is unfulfilled, the `calculator.wasm` component could not run by itself in its current form. To fulfill the `add` import, so that only `calculate` is exported, you would need to [compose the `calculator.wasm` with some `exports-add.wasm` into a single, self-contained component](../creating-and-consuming/composing.md).
```

kate-goldenring (changes requested):
Thanks for improving the flow here. Just one requested change

kate-goldenring (approved):

PR #115, opened 2024-03-18
Updated example-host to use the latest deps

minghuaw:
`wasmtime_wasi::preview2::Table` was renamed to `ResourceTable` in version 17.0.


kate-goldenring:
@minghuaw thank you for adding this; however, this is a duplicate of https://github.com/bytecodealliance/component-docs/pull/106. That change should be in today

PR #118, opened 2024-03-19
[fix] Update to use cargo-component 0.10.1 deps in tutorial

NukeManDan:
See https://github.com/bytecodealliance/cargo-component/blob/v0.10.1/example/Cargo.toml#L10-L11 for the present requirements, addresses https://bytecodealliance.zulipchat.com/#narrow/stream/206238-general/topic/Tutorial.20issues.20in.20Rust/near/427608845

Note: This does indeed work as instructed on the README, but every build yields validation errors. Happy to fix those before merging if given a hit as to why :pray: ... small chance this is a bug with cargo component?


NukeManDan:
https://github.com/bytecodealliance/cargo-component/blob/v0.10.1/example/src/lib.rs#L23C1-L23C53 Forgot this - so we now have no errors :grin:

NukeManDan:
Also curious if https://github.com/bytecodealliance/cargo-component/blob/v0.10.1/example/Cargo-component.lock is something that could be added, if that is now working? Some way to warn an incompatible version of cargo component is used at least?

rylev (approved):

kate-goldenring (approved):

kate-goldenring:
Thank you @NukeManDan for picking this up!

Issue #10, opened 2023-08-09
Create docs: Advanced Topics


kate-goldenring:
Per [proposed TOC](https://github.com/bytecodealliance/component-docs/pull/1), add section that includes:

- Binary Interface - how strings are laid out in memory, ownership, low-level nuts and bolts
  - How Types are Laid Out in Memory
  - Function Arguments and Results
  - Memory Management

_May move content from other sections here as needed_

vados-cosmonic:
Hey do we still need this? We've got an advanced section now, and a [page for the canonical abi](https://component-model.bytecodealliance.org/advanced/canonical-abi.html) -- maybe we can add more when users ask more clarifying questions/show a need?

kate-goldenring:
Closing as completed

PR #119, opened 2024-03-21
Move the canonical abi section to an advanced section

### LINKED ISSUE #112: Move the "canonical ABI" section to a "diving deeper" section
Most users of components will never need to consider the existence of the canonical ABI as the concrete mechanism by which components interact with the outside world. While an exploration of this detail of the component model is certainly useful, I think we should place this in a section that is explicitly marked as being only useful for those who want to go beyond just using components.

rylev:
Closes #112 (and potentially #10).

This creates a new book section entitled "Advanced Topics" and moves the canonical ABI section there. Most users won't need to concern themselves with the canonical ABI, so leaving this section to the end of the book seems appropriate. 

This PR also removes a couple of references to the canonical ABI where it didn't really add much (and could potentially be a source of confusion).


kate-goldenring (approved):
Thanks, @rylev. This opens to door for other advanced topics to come.

itowlson (approved):
I don't love "Advanced" as a long-term organising strategy, because it can turn into a grab-bag of unrelated topics, and how do I know if the thing I was to find out about is "advanced" or "basic."  But it's fine as a short-term measure until we better understand what sort of topics we have.

rylev:
@kate-goldenring @itowlson I don't have permission to merge, so one of you will have to.

PR #125, opened 2024-04-09
Fix typo in wasmtime.md

jahs-tech:
Fix a simple typo in the docs.

Did not perform any further validation (since I'm only on the web editor, and it seems as safe as a change can be).


itowlson (approved):
Thanks for the catch!

Issue #130, opened 2024-04-26
Incorrect link to the Components page in the Canonical ABI page


OsamaAhmad00:
[Here](https://component-model.bytecodealliance.org/advanced/canonical-abi.html), the URL of the `Components` page is set to `https://component-model.bytecodealliance.org/advanced/components.html` instead of `https://component-model.bytecodealliance.org/design/components.html`

IFcoltransG:
I believe this issue can be closed.

PR #132, opened 2024-04-27
Fix incorrect link to the Components page

OsamaAhmad00:
Fixing issue #130


itowlson (approved):
Thanks for the catch and the fix!

PR #134, opened 2024-04-30
Update to wit-bindgen-rt 0.23.0

NukeManDan:
Update and rebuild bindings using `wit-bindgen-rt 0.23.0` and `cargo-component 0.11.0`

Should fix https://bytecodealliance.zulipchat.com/#narrow/stream/206238-general/topic/Running.20component.20model.20example.20calculator/near/436159158


component-model/examples/tutorial/adder/src/bindings.rs
```diff
@@ -1,9 +1,12 @@
-// Generated by `wit-bindgen` 0.21.0. DO NOT EDIT!
+// Generated by `wit-bindgen` 0.24.0. DO NOT EDIT!
 // Options used:
```
NukeManDan, line 2: Unclear why the version reported is wrong here... Upstream oopsie?

component-model/examples/tutorial/adder/Cargo.toml
```diff
@@ -5,7 +5,7 @@ edition = "2021"
 
 [dependencies]
 bitflags = "2.4.2"
-wit-bindgen-rt = "0.20.0"
+wit-bindgen-rt = { version = "0.23.0", features = ["bitflags"] }
```
NukeManDan, line 8: see https://github.com/bytecodealliance/cargo-component/blob/920dbdc8cec4a06c5bcd04f64d3d9a5fc3f09896/example/Cargo.toml#L10

component-model/examples/tutorial/adder/src/bindings.rs
rylev: I believe that `cargo-component` uses [`wit-bindgen 0.24.0`](https://crates.io/crates/wit-bindgen/0.24.0) for code generation. I think perhaps we could use [`wit-bindgen-rt 0.24.0`](https://crates.io/crates/wit-bindgen-rt/0.24.0) instead of 0.23.0, but perhaps we should make that change to the `cargo-component` repo's example first before here.

NukeManDan: Thanks @rylev ! Updated here too now in ae223be

rylev (approved):

Issue #139, opened 2024-06-02
Can we have more examples that cover resource, record, etc.


chaosprint:

itowlson:
@chaosprint It might take us a bit to decide on how we want to organise things, but if you want to know about resources pronto, see #140.  I'll have a look at putting something similar together for records, which I hope will be a lot simpler.

PR #140, opened 2024-06-04
Working with records and resources in Rust

itowlson:
This addresses the resources side of #139.

ETA: and the records.

### DIFF
--- a/component-model/src/language-support/rust.md
+++ b/component-model/src/language-support/rust.md
@@ -276,9 +276,60 @@ $ wasmtime run ./my-composed-command.wasm
 1 + 1 = 579  # might need to go back and do some work on the calculator implementation
 ```
 
+## Using user-defined types
+
+[User-defined types](../design/wit.md#user-defined-types) map to Rust types as follows.
+
+| WIT type   | Rust binding |
+|------------|--------------|
+| `record`   | `struct` with public fields corresponding to the record fields |
+| `variant`  | `enum` with cases corresponding to the variant cases |
+| `enum`     | `enum` with cases corresponding to the enum cases, with no data attached |
+| `resource` | [See below](#using-resources) |
+| `flags`    | Opaque type supporting bit flag operations, with constants for flag values |
+
+For example, consider the following WIT:
+
+```wit
+interface types {
+    enum operation {
+        add,
+        sub,
+        mul,
+        div
+    }
+
+    record expression {
+        left: u32,
+        operation: operation,
+        right: u32
+    }
+
+    eval: func(expr: expression) -> u32;
+}
+```
+
+When exported from a component, this could be implemented as:
+
+```rust
+impl Guest for Implementation {
+    fn eval(expr: Expression) -> u32 {
+        // Record fields become public fields on a struct
+        let (l, r) = (expr.left, expr.right);
+        match expr.operation {
+            // Enum becomes an enum with only unit cases
+            Operation::Add => l + r,
+            Operation::Sub => l - r,
+            Operation::Mul => l * r,
+            Operation::Div => l / r,
+        }
+    }
+}
+```
+
 ## Using resources
 
-[Resources](../design/wit.md#resources) are handles to entities that live outside the component, for example in a host, or in a different component. You can import or export resources via `cargo component`.
+[Resources](../design/wit.md#resources) are handles to entities that live outside the component, for example in a host, or in a different component.
 
 ### Example
 
@@ -288,23 +339,23 @@ In this section, our example resource will be a [Reverse Polish Notation (RPN)](
 package docs:rpn@0.1.0;
 
 interface types {
-	enum operation {
-		add,
-		sub,
-		mul,
-		div
-	}
+    enum operation {
+        add,
+        sub,
+        mul,
+        div
+    }
 
-	resource engine {
-		constructor();
-		push-operand: func(operand: u32);
-		push-operation: func(operation: operation);
-		execute: func() -> u32;
-	}
+    resource engine {
+        constructor();
+        push-operand: func(operand: u32);
+        push-operation: func(operation: operation);
+        execute: func() -> u32;
+    }
 }
 
 world calculator {
-	export types;
+    export types;
 }
 ```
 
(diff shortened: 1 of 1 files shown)


component-model/src/language-support/rust.md
```diff
@@ -275,3 +275,263 @@ fn main() {
...
+            Operation::Add => left + right,
+            Operation::Sub => left - right,
+            Operation::Mul => left * right,
+            Operation::Div => left / right,
+        };
+        stack.push(result);
+    }
+
+    fn execute(&self) -> u32 {
+        return self.stack.borrow_mut().pop().unwrap(); // TODO: error handling!
```
rylev, line 408:
```suggestion
        self.stack.borrow_mut().pop().unwrap() // TODO: error handling!
```

component-model/src/language-support/rust.md
```diff
@@ -275,3 +275,263 @@ fn main() {
...
+```
+
+4. The resource now appears in the generated bindings as a `struct`, with appropriate associated functions. Use these to construct a test app:
+
+```rust
+#[allow(warnings)]
+mod bindings;
+
+fn main() {
+    let calc = bindings::docs::rpn::types::Engine::new();
```
rylev, line 462: nit: `use` the `Engine` and `Operation` types.

component-model/src/language-support/rust.md
```diff
@@ -275,3 +275,263 @@ fn main() {
...
+    calc.push_operand(2);
+    calc.push_operation(bindings::docs::rpn::types::Operation::Add);
+    let sum = calc.execute();
+    println!("{sum}");
+}
+```
+
+You can now build the command component and [compose it with the `.wasm` component that implements the resource.](../creating-and-consuming/composing.md). You can then run the composed command with `wasmtime run`.
+
+### Implementing and exporting a resource implementation in a host
```
rylev, line 473: I can see how this is useful, but it is very `wasmtime` specific. Should this maybe live somewhere else, and we simply point to it from here?

rylev (approved):
Looks like an improvement overall!

component-model/src/language-support/rust.md
```diff
@@ -275,3 +275,263 @@ fn main() {
...
+
+When exported from a component, this could be implemented as:
+
+```rust
+impl Guest for Implementation {
+    fn eval(expr: Expression) -> u32 {
+        // Record fields become public fields on a struct
+        let (l, r) = (expr.left, expr.right);
+        match expr.operation {
+            // Enum becomes an enum withb only unit cases
```
FrankReh, line 320: *with*

component-model/src/language-support/rust.md
```diff
@@ -275,3 +275,263 @@ fn main() {
...
+            Operation::Div => l / r,
+        }
+    }
+}
+```
+
+## Using resources
+
+[Resources](../design/wit.md#resources) are handles to entities that live outside the component, for example in a host, or in a different component. You can import or export resources via `cargo component`.
+
```
FrankReh, line 333: > "You can import or export resources via `cargo component`.

Is there a more  precise way to convey this idea? All other discussion I've seen about Wasm Components reserves the terms import and export for things that are done either by a compiler or a runtime. This kind of makes it sound like it is also done at the OS level. Something like "The `new` and `build` subcommands of `cargo component`, as shown above, can be used to create a Wasm component binary that uses `resources`, either to import, export or both."

That's not great either.

component-model/src/language-support/rust.md
```diff
@@ -275,3 +275,263 @@ fn main() {
...
+    stack: RefCell<Vec<u32>>,
+}
+```
+
+> Why is the stack wrapped in a `RefCell`? As we will see, the generated Rust trait for the calculator engine has _immutable_ references to `self`. But our implementation of that trait will need to mutate the stack. So we need a type that allows for interior mutability, such as `RefCell<T>` or `Arc<RwLock<T>>`.
+
+3. The generated bindings for an exported resource include a trait named `GuestX`, where `X` is the resource name. For the calculator `engine` resource, the trait is `GuestEngine`. Implement this trait on the `struct` from step 2:
+
+```rust
+impl GuestEngine for CalcEngine {
```
FrankReh, line 383: Could this show the `use` statement, where the `GuestEngine` is defined? Maybe I missed it - so many other aspects of this description are so good.

component-model/src/language-support/rust.md
```diff
@@ -275,3 +275,263 @@ fn main() {
...
+```rust
+struct Implementation;
+impl Guest for Implementation {
+    type Engine = CalcEngine;
+}
+
+export!(Implementation);
+```
+
+This completes the implementation of the calculator `engine` resource. Run `cargo component build` to create a component `.wasm` file.
```
FrankReh, line 424: Would it be worth pointing out that this build step also creates/recreates the bindings.rs module based on the .wit file(s)? I guess maybe that's overkill but I don't find bindings.rs mentioned in this rust.md file.

component-model/src/language-support/rust.md
```diff
@@ -275,3 +275,263 @@ fn main() {
...
+
+### Importing and consuming a resource in a component
+
+To use the calculator engine in another component, that component must import the resource.
+
+1. Create a command component as shown in previous sections.
+
+1. Add a `wit/world.wit` to your project, and write a WIT world that imports the RPN calculator types:
```
FrankReh, line 432: 1 -> 2

component-model/src/language-support/rust.md
FrankReh: If there is no other place right now, could this section be kept here with a disclaimer that this is in fact *wasmtime* specific and will be moved once there is an agreed upon place or places for Wasm Component Host examples?

FrankReh:
This is super useful!

component-model/src/language-support/rust.md
itowlson: @rylev Agree, that would be ideal.  I could offer it to the [wasmtime book](https://docs.wasmtime.dev/) if you think they would be interested?  It seems at a slightly different level from a lot of what's in there so not sure.  I had to puzzle it out by following compiler errors and asking Wasmtime developers, so I'm reluctant to discard it...!

component-model/src/language-support/rust.md
itowlson: I think we can just drop the problematic sentence.

component-model/src/language-support/rust.md
itowlson: In this case there is no `bindings.rs` - the bindings are generated by macro-expansion of the `wit_bindgen::generate!(...)` which is created as part of `cargo component new`.  I didn't go over that because it's not specific to resources.

rylev: Wait - doesn't `cargo component new` use a generated `bindings.rs` file and not `wit_bindgen`?

itowlson: @rylev Bother, yes. I've been working off a sample which was _not_, in fact generated by `cargo component` and I had forgotten where it came from.  What an idiot.  I will recreate the sample and rework this section.  Thank you both for catching this.

itowlson: Okay this is now recreated and I have added the mention of using `cargo component build` to regenerate `bindings.rs`.  Again, apologies and thanks both for the catch.

component-model/src/language-support/rust.md
rylev: I'm fine with keeping it for now if we can track somewhere that the plan is to move this content and link to it in the future.

rylev (approved):

Issue #38, opened 2023-09-07
Language Guide: Add C


kate-goldenring:
Add  instructions for building C modules and componentizing them

PR #66, opened 2023-09-10
Add initial documentation for C tooling

seanisom:
Fixes #38 

Creating as a draft for now as I would like to see #44 merged first as that will change the structure of `/language-support` - I will update the table of contents to match that once I can rebase from that.

PR #147, opened 2024-08-25
Add initial documentation for C tooling [continued]

rajsite:
A continuation of https://github.com/bytecodealliance/component-docs/pull/66 to add initial C/C++ documentation.

- Merged latest changes from main
- Updates links to C/C++ language support page
- Updates the example cli output with the results from a local manual run
- Validated the instructions with latest release tooling (`wasi-sdk-24.0-x86_64-windows.tar.gz`, `wasm-tools-1.216.0-x86_64-windows.zip`, `wit-bindgen-0.30.0-x86_64-windows.zip`)

Fixes https://github.com/bytecodealliance/component-docs/issues/38


rajsite:
Doesn't look like owners are defined for reviews so just in case, FYI @seanisom @kate-goldenring

component-model/src/language-support/c.md
```diff
@@ -0,0 +1,106 @@
...
+
+int32_t example_add(int32_t x, int32_t y)
+{
+	return x + y;
+}
+```
+
+Now, you can compile the function into a Wasm module via clang:
+```sh
+clang add.c example.c example_component_type.o -o add-core.wasm -mexec-model=reactor
```
kate-goldenring, line 35: Can you add a note to replace `clang` with the appropriate setup from above for the `wasi-sdk`, ensuring it is pointing to the local `clang` from `wasi-sdk`.

kate-goldenring (approved):
LGTM! Thank you for pushing this to the finish line. I left one note that could help make it a little smoother.

seanisom (approved):
LGTM! Thanks for the help

component-model/src/language-support/c.md
rajsite: To be precise are you referencing the instructions in `wasi-sdk` that describe running the following:

```
export WASI_SDK_PATH=`pwd`/wasi-sdk-${WASI_VERSION_FULL}
CC="${WASI_SDK_PATH}/bin/clang --sysroot=${WASI_SDK_PATH}/share/wasi-sysroot"
$CC foo.c -o foo.wasm
```

Which would result in the suggestion:

```suggestion
$CC add.c example.c example_component_type.o -o add-core.wasm -mexec-model=reactor
```

I was concerned about that approach because:
- Those instructions aren't really windows friendly (the wasi-sdk docs seem to poorly discuss cross-platform instructions).
- It masks the specific tooling used, i.e. the `clang` cli (I won't claim to be an expert here at all but my experience is that the `$CC` convention is more of a Makefile convention, but I'll absolutely cede to any other opinion).

What about an alternate suggestion stating:

```
Note: Use the `clang` included in the wasi-sdk installation, for example at `<WASI_SDK_PATH>/bin/clang`
```

I also feel more strongly about getting something up and doing iterative changes after so if you have a specific alternatives lets go with that!

kate-goldenring: Yes, that was it. Thanks for laying it all out. I like your suggested note and agree getting something up is priority. We can merge this today, whether or not the note is added

rajsite: @kate-goldenring added the note as a markdown quote (the pattern I see for other `Note:` comments). Let me know if any other tweaks otherwise ready to go!

seanisom: I tend to agree -  I originally worded as such since I was on Windows when I first wrote this. For me it was easier to use one of the already-installed `clang`s and link appropriately to wasi-sdk artifacts with the lack of symlinks and makefile conventions. A clarification here to prefer the wasi-sdk clang makes sense!

PR #149, opened 2024-08-28
Switch to a consistent note format and code snippet indentation level

rajsite:
Made the following changes:

- Used a consistent note style aligning with what seemed to be the most common (prefixing a quote block with the `ⓘ` character)
- Made code blocks aligned to their indentation level in lists
- Used inline-code styling on the wit page consistently when referencing a type (did not evaluate other pages)


itowlson:
Could we settle on `>` instead of `> (i symbol)`?  To me the extra `i` is noisier, is a faff to type, and the symbol comes out very spindly.  I'm okay with it if there's a strong reason, of course!

rajsite:
> Could we settle on `>` instead of `> (i symbol)`? To me the extra `i` is noisier, is a faff to type, and the symbol comes out very spindly. I'm okay with it if there's a strong reason, of course!

I like the visual queue of  ⓘ as an icon but the implementation as text I wasn't a huge fan of. Would be nicer as a custom template in mdbook and as an icon image instead of text if desired in the future.

Went ahead and pushed a change to remove it for now.

rajsite:
@itowlson Thanks for catching the bad find-replaces 😅
Also made a few tweaks to the other examples:
- whitespace between lines with comments
- switched to just normal comments from doc comments (the existing examples didn't look like good representative doc comments but up for other suggestions)
- some other comment alignment and manual wrapping

itowlson (approved):
Brilliant, thanks for this and for your attention to detail!

Issue #157, opened 2024-09-27
Enable site analytics via Plausible.io


disquisitioner:
SIG-Documentation would like enable site analytics and usage tracking for the online [Component Model book](https://component-model.bytecodealliance.org/) using the Bytecode Alliance's existing Plausible.io account (which we use today for bytecodealliance.org).  To do that we need to embed the following script in the `<head>` section of every page on the site, which for the Component Model book means properly via mdBook.

`<script defer data-domain="component-model.bytecodealliance.org" src="https://plausible.io/js/script.js"></script>`

I can handle the rest of the setup within our Plausible.io account once the script is inserted.  Additional helpful information about the Plausible.io tracking snippet can be found [here](https://plausible.io/docs/plausible-script).

kate-goldenring:
@disquisitioner just added the header. Let me know if that pipes through on your side.

disquisitioner:
Looks good!  Plausible.io ran its verifier to confirm the snippet was present and everything looked OK, so we should be all set.  It may take a little while for data to show up in the analytics page for component-model.bytecodealliance.org, so I'll keep an eye out there and let you know.

Thanks!

kate-goldenring:
@disquisitioner can you remind me how we get access to the analytics? Is there a link to the stats that you can share

kate-goldenring:
On second look, I think for the component docs book, these analytics could be shared publicly. We could create a google document that you could (if willing) periodically paste the snapshots into. Alternatively, we may be able to use Plausible's [stats api](https://plausible.io/docs/stats-api) to run a cron job in CI monthly to produce a report.

vados-cosmonic:
Thanks Kate! So I really wanted to know the most visited pages, how much time people spend on particular pages, and stuff like that. Basically, if 99% of the time people check out a particular language guide, I think it would make sense to spend more time making sure that particular language guide is effective.

> Alternatively, we may be able to use Plausible's [stats api](https://plausible.io/docs/stats-api) to run a cron job in CI monthly to produce a report.

This would be fantastic.

kate-goldenring:
@disquisitioner sorry for the delay here. 

> Would it be better for me to publish the info to a Zulip stream, e.g. a topic in #SIG-Documentation?

That sounds like a good idea, and it provides an easy way to immediately discuss the analytics

[...]

PR #160, opened 2024-10-01
Add header to support plausible analytics

kate-goldenring:
fixes https://github.com/bytecodealliance/component-docs/issues/157 (i hope)

I verified that it updates the html of the pages to add the script in the headers


disquisitioner:
I used the Plausible.io admin dashboard to verify that the snippet is detected properly, and now have a analytics page there for component-model.bytecodealliance.org.  I presume it'll take a little while for analytics data to accumulate so will watch for that to happen.

Thanks!

PR #163, opened 2024-10-09
fix: use wac plug in tutorial and running guide

kate-goldenring:
a few more spots where we could be plugging


itowlson (approved):
PLUG ALL THE THINGS

PR #170, opened 2024-11-05
go: update to use the new wit-bindgen-go tool and wasip2 target of TinyGo

Mossaka:


component-model/src/language-support/go.md
```diff
@@ -1,288 +1,178 @@
 # Go Tooling
 
-The [TinyGo toolchain](https://tinygo.org/docs/guides/webassembly/wasi/) has native support for WASI
-and can build Wasm core modules. With the help of some component model tooling, we can then take
-that core module and embed it in a component. To demonstrate how to use the tooling, this guide
-walks through building a component that implements the `example` world defined in the [`add.wit`
+The [TinyGo compiler](https://tinygo.org/) v0.33.0 and above has native support for the WebAssembly Component Model and WASI 0.2.0. This guide walks through building a component that implements `example` world defined in the [`add.wit`
```
ydnar, line 3: While technically TinyGo 0.33.0 has support for `wasip2`, 0.34.0 adds support for `go:wasmexport` and has a number of quality of life fixes for WASI that we might want to say "use 0.34.0" here.

ydnar (approved):

component-model/src/language-support/go.md
```diff
@@ -1,288 +1,178 @@
...
-
-For this example, we will use the following world, which moves the add function behind an `add` interface:
+The `wasip2` target of TinyGo requires the imports of `wasi:cli/imports@0.2.0` and thus we need to include them in the `add.wit`. Below is the minimal `add.wit` file that includes the required imports:
 
 ```wit
 package docs:adder@0.1.0;
 
-interface add {
-    add: func(a: u32, b: u32) -> u32;
+world adder {
```
kate-goldenring, line 24: Can you add a note as to why you needed to add other WASI dependencies to the adder world?

component-model/src/language-support/go.md
```diff
@@ -1,288 +1,178 @@
...
 func main() {}
 ```
 
-After implementing the adder world, we need to load it by passing it to the `SetExportsDocsAdder0_1_0_Add`
-function from our bindings (`adder.go`). Since our component is a library, `main` will not be called. However, only Go
-programs with `main` can target WASI currently. As a loophole, we will initialize our `AdderImpl`
-type inside an `init` function. Go's `init` functions are used to do initialization tasks that
+Go's `init` functions are used to do initialization tasks that
 should be done before any other tasks. In this case, we are using it to export the `Add` function and
 make it callable using the generated C bindings (`adder.c`). After populating the `init` function,
```
kate-goldenring, line 138: Looks like c bindings are no longer generated/needed. Should this be updated?

component-model/src/language-support/go.md
```diff
@@ -1,288 +1,178 @@
...
-// `init` function which is always called first when a Go package is run.
-func init() {
-	example := AdderImpl{}
-	SetExportsDocsAdder0_1_0_Add(example)
-}
-
-// main is required for the `WASI` target, even if it isn't used.
-func main() {}
-```
+TinyGo will invoke `wasm-tools` to embed the WIT file to the module and componentize it.
```
kate-goldenring, line 144: So this is why wasm-tools is required? This may make it a little clearer 
```suggestion
Under the hood, TinyGo invokes `wasm-tools` to embed the WIT file to the module and componentize it.
```

component-model/src/language-support/go.md
```diff
@@ -1,288 +1,178 @@
...
-export COMPONENT_ADAPTER_REACTOR=/path/to/wasi_snapshot_preview1.reactor.wasm
-tinygo build -o add.wasm -target=wasi add.go
-wasm-tools component embed --world adder ./add.wit add.wasm -o add.embed.wasm
-wasm-tools component new -o add.component.wasm --adapt wasi_snapshot_preview1="$COMPONENT_ADAPTER_REACTOR" add.embed.wasm
+$ tinygo build -target=wasip2 -o add.wasm --wit-package add.wit --wit-world adder main.go
 ```
 
-We now have an add component that satisfies our `adder` world, exporting the `add` function, which
-we can confirm using the `wasm-tools component wit` command:
+We can confirm using the `wasm-tools component wit` command:
```
kate-goldenring, line 150: Can we keep the more verbose wording?

```suggestion
We now have an add component that satisfies our `adder` world, exporting the `add` function, which
we can confirm using the `wasm-tools component wit` command:
```

component-model/src/language-support/go.md
```diff
@@ -1,288 +1,178 @@
...
 
 ```sh
```
kate-goldenring, line 152: nit: can you use console when displaying both command and output? I know it wasn't doing this properly previously either
```suggestion
```console
```

kate-goldenring (commented):
Thank you for updating this @Mossaka. Just some clarifiers/suggestions

component-model/src/language-support/go.md
Mossaka: Done

kate-goldenring (approved):

PR #175, opened 2024-11-15
component-model/src/language-support/go: fix go docs

Mossaka:
* Delete a confusing where I described an alternative way to write the WIT package
* Fixed the file name of the resolved WIT package


kate-goldenring:
@Mossaka does this resolve a specific issue that we should close too?

kate-goldenring (approved):

PR #186, opened 2024-12-11
fix a few `wasi:cli` typos

liigo:


kate-goldenring (approved):
Thank you for fixing this!

PR #191, opened 2025-01-15
remove redundant wording from python.md

dicej:
This is a follow up to https://github.com/bytecodealliance/component-docs/pull/190, which included some redundant wording.

### DIFF
--- a/component-model/src/language-support/python.md
+++ b/component-model/src/language-support/python.md
@@ -100,7 +100,7 @@ Component built successfully
 Wasm components can also be invoked from Python applications. This walks through using tooling
 to call the [`app.wasm` component from the examples](../../examples/example-host/add.wasm).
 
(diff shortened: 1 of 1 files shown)


itowlson (commented):
For me, the original words give a stronger sense of "it's the tool injecting these imports."  If other folks prefer the more concise wording, though, then I'm happy to go with it.

dicej:
I think the current wording is grammatically incorrect (mea cupla).  Specifically: "...which components built with `componentize-py` use by default by generating components which...".  When you ignore the middle bits, it reads "...which components ... use by default by generating components which..."; but the components aren't generating components -- `componentize-py` is.  So I think something needs to change, at least.

dicej:
> How about:
> 
> > This is because `wasmtime-py` does not yet support [resources](../design/wit.md#resources), and `componentize-py` by default causes components to use resources by importing most of the `wasi:cli` world.
> 
> Sentence structure is a bit fiddly though...!
> 
> > This is because `wasmtime-py` does not yet support [resources](../design/wit.md#resources), and when you run `componentize-py` _without_ `--stub-wasi`, it imports resources from the `wasi:cli` world.  _[or "it generates components which import resources..."?]_

Either one works for me.  I do prefer "generates components which import resources" over "causes components to use resources".  I'll push an update shortly.

dicej:
How does this sound (per my latest push)?

> This is because `wasmtime-py` does not yet support [resources](../design/wit.md#resources), and `componentize-py` by default generates components which use resources from the `wasi:cli` world.

itowlson (approved):
Thanks for your patience!

dicej:
Thanks for the thoughtful review!

Issue #24, opened 2023-08-24
Update language guide `add` example to use a world with an interface


kate-goldenring:
To better align with tutorial, update wit and examples to:
```wit
interface add {
    add: func(a: u32, b: u32) -> u32
}

world adder {
    export add
}
```

kate-goldenring:
I started looking into this but i must say that i am stumped by how to implement a component with an interface for `jco` and `componentize-py`. I am thinking of waiting till after Componentize the World on Sept 8th, as there will be explicit tutorials from the language experts that we can learn from and encorporate

vados-cosmonic:
Hey @kate-goldenring is this still a priority? I do agree that we should use an interface (`adder`?) instead of the bare function export (and at this point it's pretty well supported!) -- so I could work on this and update all the docs and code samples *OR* we could leave it as a bare function for now :)

kate-goldenring:
I think this is something we should still do, so I don't think we should close this. I personally don't have the time to take it on but i think it would be great to get to. Potentially in a way where all examples just implement the [`adder`](https://github.com/bytecodealliance/component-docs/blob/main/component-model/examples/tutorial/wit/adder/world.wit) world from the tutorial instead of having a separate add interface for the example host which causes a lot of confusion, which we observed in the JS lang docs revamp.

PR #192, opened 2025-01-16
fix: use adder interface for host

vados-cosmonic:
This PR updates the host to use the existing `adder` interface, and actually re-use the existing WIT file so it *should* be harder to diverge in the future.

Resolves https://github.com/bytecodealliance/component-docs/issues/24


kate-goldenring:
@vados-cosmonic many of the language guides reference this example world (`add.wit). This PR will need to update all those guides to use the `adder` world, as well, before we can delete that file and update the example host

kate-goldenring:
@vados-cosmonic i'd rather move everything to the `adder` world (which has an interface) as updating the language guides to use a world with an interface is the explicit goal of the issue this PR aims to resolve: https://github.com/bytecodealliance/component-docs/issues/24

The c guide takes into account the lack of interface because our example has no interface not because it should -- I'd rather it does not discuss using worlds that don't use interfaces as it is a best practice to use interfaces IMO.

vados-cosmonic:
Alright, that sounds good -- we'll knock out #24 as well at the same time in this PR!

> The c guide takes into account the lack of interface because our example has no interface not because it should -- I'd rather it does not discuss using worlds that don't use interfaces as it is a best practice to use interfaces IMO.

Absolutely agree here -- while the simple function case might be good for simplicity (less new keywords mentioned, etc), the best practice is definitely using interfaces.

kate-goldenring:
@vados-cosmonic just a check on this, i know it is a lot of updates. I can also help take some of this on as well. Would be great to get rid of the interface-less add.wit interface and only use the adder world

vados-cosmonic:
Hey thanks for the ping -- I got busy and forgot about this -- need to get back to it! I'll  find time to wrap this up by next week!

component-model/examples/tutorial/wit/adder/world.wit
```diff
@@ -1,7 +1,7 @@
 package docs:adder@0.1.0;
 
 interface add {
-    add: func(a: u32, b: u32) -> u32;
+    add: func(x: u32, y: u32) -> u32;
```
kate-goldenring, line 4: any reason for the variable name changes? If you want to change them, can you change all references in the language guides to match?

component-model/src/language-support/c.md
```diff
@@ -1,83 +1,142 @@
...
+wasm-tools component new adder.wasm --adapt wasi_snapshot_preview1.wasm -o adder.component.wasm
 ```
 
+## 7. Inspect the built component
+
 Finally, you can inspect the embedded wit to see your component (including any WASI imports if necessary):
-```sh
->wasm-tools component wit add-component.wasm
+
+```
```
kate-goldenring, line 138:
```suggestion
```console
```

component-model/src/language-support/python.md
```diff
@@ -97,8 +97,8 @@ Component built successfully
 
 ## Running components from Python Applications
 
-Wasm components can also be invoked from Python applications. This walks through using tooling
-to call the [`app.wasm` component from the examples](../../examples/example-host/add.wasm).
+Wasm components can also be invoked from Python applications. This section walks through using tooling
+to call the [pre-built `app.wasm` component][add-wasm] in the examples.
```
kate-goldenring, line 101: Looks like this does not exist. Could you add it in this PR or we can follow up with it.

kate-goldenring (approved):
LGTM! Thanks @vados-cosmonic. I left a few nits that could be addressed in this PR or a subsequent one

component-model/examples/tutorial/wit/adder/world.wit
vados-cosmonic: Hey so this was actually to make things consistent -- and you're absolutely right -- I actually was in the process of going through *all* the language guides, almost through with C# (I don't run Windows so that ones  a bit more challenging).

I'm redoing the guides and making sure they're working/up to date locally first -- I probably should have moved this back to draft or something!

component-model/src/language-support/c.md
vados-cosmonic: Ah so this was intentional -- while `sh` isn't right because this isn't a script (IMO), `console` isn't right either because this isn't just a command -- this is a command with it's output. For those blocks I normally just leave it empty.

component-model/src/language-support/python.md
vados-cosmonic: Ah good catch, will fix! Python I haven't finished going back through either yet.

vados-cosmonic: Fixed this link!

component-model/examples/tutorial/wit/adder/world.wit
vados-cosmonic: OK, went through and made this consistent!

vados-cosmonic:
Hey @kate-goldenring finally this is done and I think mergable! It should merge *after* James's awesome work here: https://github.com/bytecodealliance/component-docs/pull/221 (I already merged his changes in actually but just don't want to "steal" his hard work!)

Definitely happy to follow with follow ups as well if you find anything I missed -- I went through all the guides for all the languages and they all worked just fine. There may be some step numbering issues (I introduced step numbering where there wasn't for many of the guides).

kate-goldenring:
@vados-cosmonic sounds good! Can you rebase and we should be set to merge?

vados-cosmonic:
Hey @kate-goldenring apologies for the delay here -- should be good to go!

PR #195, opened 2025-01-30
Fix Go example

rellfy:
I made some changes to the Go example because I had issues running it in its current state:

1. Looks like the instructions for running `wit-bindgen-go` are duplicated (and the 1st one doesn't work because no project was created at the point)
2. Update `main.go` to match example values
3. Fix `--wit-package` argument to `tinygo`


kate-goldenring (approved):

Mossaka:
LGTM! Thanks

PR #211, opened 2025-03-10
Update artifact action maybe correctly this time

itowlson:
Sorry Kate I misread the error message as "you're using v3 and must upgrade (to v4)" when it was actually saying "one of your things uses v3 internally so you must upgrade _that_ (as it turns out, to v3)."  Sorry for the churn.


kate-goldenring (approved):
V3 for the win!

PR #217, opened 2025-03-16
C# example - don't confuse int/s32 with uint/u32

MendyBerger:
The C# examples aren't consistent with converting s32 to int and u32 to uint, leading to a broken example.
For consistency and simplicity, I've gone ahead and changed all numbers to only use unsigned integers.


kate-goldenring (approved):
LGTM. Thanks @MendyBerger

PR #218, opened 2025-03-18
docs: Clarify setup for "Rust Components" tutorial

mkatychev:
This PR aims to make explicit some assumptions made in the Rust Components tutorial.

Changes made to `component-model/src/language-support/rust.md`:
* Added _Prerequisites_ section
* Added link to WIT package naming and used explicit package/interface/world names such as `example:component/example` to help disambiguate between similar sounding names such as `add` and `adder`
* Certain code is now inlined for consistency: https://github.com/bytecodealliance/component-docs/blob/2e207e41d470bca94b3f307fb5a9fc77e797bb71/component-model/src/language-support/rust.md?plain=1#L51-L53
* Fixed incorrect filepath in _Referencing the package to import_ section
* Removed some leading `$` from shell commands to make them [easier to copy](https://github.com/rust-lang/mdBook/issues/864)


component-model/src/language-support/rust.md
```diff
@@ -77,41 +124,30 @@ $ cargo run --release -- 1 2 ../add/target/wasm32-wasip1/release/add.wasm
...
-    }
-}
+{{#include ../../examples/tutorial/adder/src/lib.rs}}
 ```
 
 ## Importing an interface with `cargo component`
 
 The world file (`wit/world.wit`) generated for you by `cargo component new --lib` doesn't specify any imports.
 
-> `cargo component build`, by default, uses the Rust `wasm32-wasi` target, and therefore automatically imports any required WASI interfaces - no action is needed from you to import these. This section is about importing custom WIT interfaces from library components.
```
mkatychev, line 114: Mentioning the compilation target too much info for someone just going through the tutorial.

component-model/src/language-support/rust.md
```diff
@@ -1,59 +1,106 @@
...
+
 struct Component;
 
 impl Guest for Component {
     fn add(x: i32, y: i32) -> i32 {
         x + y
     }
 }
+
+bindings::export!(Component with_types_in bindings);
```
mkatychev, line 86: excluding the `bindings::export!` and `mod bindings` section made it difficult to follow along sometimes.

component-model/src/language-support/rust.md
```diff
@@ -77,41 +124,30 @@ $ cargo run --release -- 1 2 ../add/target/wasm32-wasip1/release/add.wasm
...
-use bindings::exports::docs::adder::add::Guest;
-
-struct Component;
-
-impl Guest for Component {
-    fn add(a: u32, b: u32) -> u32 {
-        a + b
-    }
-}
+{{#include ../../examples/tutorial/adder/src/lib.rs}}
```
mkatychev, line 140: Not against reverting the `#include` here, the tradeoff here is that the raw markdown is less readable but this will likely result in less code drift overall.

vados-cosmonic:
Hey @mkatychev would you mind taking another look at this? One of the thing we've wanted to do for a bit was refactor to use the *same* WIT files everywhere, and we finally have that in now for all the guides -- I think the changes here are in conflict with that, but it *should* simplify your diff as well.

component-model/src/language-support/rust.md
```diff
@@ -1,99 +1,103 @@
...
 ```
 
-`cargo component bindings` will generate bindings for the world specified in a package's `Cargo.toml`. In particular,
-`cargo component` will create a `Guest` trait that a component should implement.
+`cargo-component` will generate bindings for our
+world and create a `Guest` trait that a component should
+implement
 
-## 5. Implement the generated `Guest` trait
+## Implementing the `Guest` trait
```
mkatychev, line 75: @vados-cosmonic I've removed the numbering for the headers to simplify anchor link generation, happy to add them back in.

component-model/src/language-support/rust.md
```diff
@@ -1,99 +1,103 @@
...
 
-Rust has first-class support for the component model via [the `cargo component` tool](https://github.com/bytecodealliance/cargo-component).
+Rust has first-class support for the component model via the [`cargo-component`
+tool][cargo-component]. We will be using
+the `cargo component` subcommand to create WebAssembly components using Rust as
+the component's implementation language.
 
-`cargo component` is is a `cargo` subcommand for creating WebAssembly components
-using Rust as the component's implementation language.
+> [!NOTE]
```
mkatychev, line 8: https://github.com/Michael-F-Bryan/mdbook-linkcheck/issues/93 this is going to produce warnings until https://github.com/Michael-F-Bryan/mdbook-linkcheck/issues/93 is addressed.

component-model/src/language-support/rust.md
```diff
@@ -1,99 +1,103 @@
...
+{{#include ../../examples/tutorial/adder/src/lib.rs}}
 ```
 
-## 6. Build the component
+## Building a Component
 
-Now, use `cargo component` to build the component, being sure to optimize with a release build.
+Now, let's build our component, being sure to optimize with a release build:
 
-```console
```
mkatychev, line 89: `console` langtype seems to produce less syntax highlighting than `sh`.

vados-cosmonic: yeah so my stance on that is that writing into a console is not normally syntax highlighted either, so it matches.

If we have a `sh`ell script, or a `bash` script, then it makes sense to use those, but my stance here is that if it's a console command you're supposed to enter there are other ways to make it more readable:

```
something \
    --like this \
    --rather-than console
```

and if it's code, *then* it should be highlighted.

component-model/src/language-support/rust.md
vados-cosmonic: I'd love feedback from @kate-goldenring here -- I think it's mostly a style thing. I think numbers help people refer to steps (ex. "please see step 3") but I could definitely go the other way. 

That said, if we want to remove the numbers here, we should remove them everywhere (which isn't a hard change either, happy to help here or in a follow up if you'd rather not do it). 

Thanks for working on this! :bow:

mkatychev: Sounds good, I was mostly following style changes off of the rust book, including using present continuous tense for titles, ex: building v.s. build.

component-model/src/language-support/rust.md
```diff
@@ -120,46 +124,31 @@ $ cargo run --release -- 1 2 ../add/target/wasm32-wasip1/release/add.wasm
...
-interface add {
-    add: func(x: u32, y: u32) -> u32;
-}
+## Exporting an interface
 
-world adder {
-    export add;
-}
-```
+Notice how our `example` world currently exports `add` as a function. It's
```
mkatychev, line 129: The world doesn't actually export `add`, this is a mistake. Will remove.

component-model/src/language-support/rust.md
vados-cosmonic: Yup, I don't think there are many ways around this for now. 

There is a way around it though, I'll put up a PR for that

component-model/src/language-support/rust.md
vados-cosmonic: Yeah if you could expand on that to note that it exports `add` as part of the `add` interface that would be great.

component-model/src/language-support/rust.md
vados-cosmonic: https://github.com/bytecodealliance/component-docs/pull/224

Feel free to cherrypick that change in here, and if this gets merged first I'll remove the other.

component-model/src/language-support/rust.md
vados-cosmonic: Yeah this makes sense... Maybe what we want is actually demarcation at a higher level (ex. in [the Rust book](https://doc.rust-lang.org/stable/book/ch01-01-installation.html) the sections are clearly demarcated in the ToC on the left), then we can go with continuous tense and still be able to refer to things?

Happy either way here, I think the Rust book is an excellent example to follow so maybe we should change to look more like them, even if I'm partial to "steps".

mkatychev: Would you mind elaborating in your 'including demarcation' comment?

vados-cosmonic: Ah I was referring to this:

image

So when someone wants to mention a specific section they can refer "Section 3.3"

component-model/src/language-support/rust.md
mkatychev: Have you looked into [`mdbook-admonish`](https://github.com/tommilligan/mdbook-admonish) for this role? [Tree-sitter documentation uses it](https://tree-sitter.github.io/tree-sitter/3-syntax-highlighting.html#admonition-output-1) and stylistically it's my personal preference, the callout sections are just codeblocks and code outputs generate their own anchor links.

vados-cosmonic: Ah so the reason I went with `mdbook-alerts` here was because the syntax matches Github, so we get good alert boxes whether in Github or on the web -- does `mdbook-admonish` support that?

mkatychev: The link above shows an alert-styled output box  (https://tree-sitter.github.io/tree-sitter/3-syntax-highlighting.html#admonition-output-1)
but all the equivalents are present https://tree-sitter.github.io/tree-sitter/3-syntax-highlighting.html#admonition-from-the-sublime-text-docs
However they use the fenced codeblock approach which works in the editor but to your point does not support quoted markdown:
https://github.com/tommilligan/mdbook-admonish/issues/72

vados-cosmonic: Yeah, `mdbook-alerts` seems "good enough" for both the web and Github display, but IMO this is worth bringing up so that everyone can discuss -- would you mind making an issue for this?

component-model/src/language-support/rust.md
mkatychev: I looked into this a bit and it looks like adding subchapters based on anchor links is still an open issue:
https://github.com/rust-lang/mdBook/issues/167

the component docs currently use numbered subchapters, however they're currently broken up by language:
Screenshot 2025-03-25 at 12 19 50

We could break up the subchapters into their own markdown page so for example `rust-scaffolding-component.md` would become:
>  _8.6.1_ Scaffolding a Component

component-model/src/language-support/rust.md
mkatychev: I've done it in this PR: https://github.com/bytecodealliance/component-docs/pull/232

component-model/src/language-support/rust.md
mkatychev: Another problem with having numbered headers is that any external references to the section (such as https://component-model.bytecodealliance.org/language-support/rust.html#6-build-the-component) will become broken if the section gets reordered. I wonder if there's a cleaner way of doing header numbering in the same way as numbered lists where the numbering gets automatically reordered:
```
1. installing
1. scaffolding
1. building
1. running
```
Codeblock above produces this:
1. installing
1. scaffolding
1. building
1. running

component-model/examples/tutorial/adder/src/lib.rs
```diff
@@ -1,13 +1,14 @@
 #[allow(warnings)]
 mod bindings;
 
+// Separating out the interface puts it in a sub-module
```
vados-cosmonic, line 4: I'm not sure this comment makes sense, because someone coming to this would probably think "as opposed to what?" and we don't have an example of exporting the function directly anymore. 

Maybe this comment could be expanded to a more general statement about bindings pathing (explaining the module hierarchy a little bit -- `exports` because it's an export, `docs` being the namepace, `adder` being the package, etc)

component-model/src/language-support/rust.md
vados-cosmonic: Hmnnn OK, rather than make any large changes in this PR, let's maintain consistency at least then discuss it -- It looks like there are a bunch of options here, and we're probably not the only two people who want to chat it out!

component-model/src/language-support/rust.md
```diff
@@ -1,99 +1,103 @@
...
 
-## 4. Generate bindings for our component
+## Generating bindings
 
-After performing these changes, we can re-generate bindings with `cargo component bindings`:
+Now that we've updated our `world.wit` and `Cargo.toml`, we can re-generate
+bindings with the command below:
 
-```console
+```sh
```
vados-cosmonic, line 67: `console` here -- I believe this was the change you made in the other PR as well

component-model/src/language-support/rust.md
```diff
@@ -1,99 +1,103 @@
...
-    add: func(x: u32, y: u32) -> u32;
-}
+## Adding the WIT world
 
-world adder {
-    export add;
-}
+We now need to change our generated `wit/world.wit` to match `docs:adder`:
+```wit
+{{#include ../../examples/tutorial/wit/adder/world.wit}}
```
vados-cosmonic, line 51: Fantastic, this will save a ton of work!

component-model/src/language-support/rust.md
```diff
@@ -1,99 +1,103 @@
...
-Now, use `cargo component` to build the component, being sure to optimize with a release build.
+Now, let's build our component, being sure to optimize with a release build:
 
-```console
+```sh
 cargo component build --release
 ```
 
-You can use `wasm-tools component wit` to output the WIT package of the component:
+You can use `wasm-tools` to inspect the WIT package generated by `cargo-component`:
```
vados-cosmonic, line 92: This isn't quite right -- we're actually inspecting the built component itself, which *happens to contain* the WIT that was used to create it. I think the old wording was right here.

component-model/src/language-support/rust.md
```diff
@@ -106,10 +110,10 @@ package docs:adder@0.1.0 {
 }
 ```
 
-### Running a Component from Rust Applications
+### Running a Component
 
 To verify that our component works, lets run it from a Rust application that knows how to run a
-component targeting the [`adder` world][adder-wit].
+component targeting the [`docs:adder/adder`](#adding-the-wit-world) world.
```
vados-cosmonic, line 116: I'm not sure being explicit about the namespace and package herehelps -- not everyone knows the more qualified naming schemes for WIT interfaces. Someone might think "what is the `docs:adder/adder` world?", and in the linked section we refer to it just as `docs:adder` (no mention of the world).

I think we want people to think in terms of worlds, which import/export interfaces -- the namespace and package are much less important IMO.

component-model/src/language-support/rust.md
vados-cosmonic: I like the includes -- certainly saves us time :+1:

component-model/src/language-support/rust.md
mkatychev: I've reverted to the numbering but kept the continuous tense here, I feel that part is a net positive.

component-model/examples/tutorial/adder/src/lib.rs
mkatychev: I've added comments to correlate the module URI with their adder `world.wit` analogues. 
Tried to keep it succinct, let me know what you think.

component-model/src/language-support/rust.md
```diff
@@ -108,10 +110,12 @@ package docs:adder@0.1.0 {
 }
 ```
 
-### Running a Component from Rust Applications
+> **WARNING:** Building with `--release` removes all debug-related information from the resulting .wasm file. When prototyping or testing locally, you might want to avoid `--release` to obtain useful backtraces in case of errors (for example, with `wasmtime::WasmBacktraceDetails::Enable`). Note: the resulting .wasm file will be considerably larger (likely 4MB+).
```
vados-cosmonic, line 113:
```suggestion
> [!WARNING]
> Building with `--release` removes all debug-related information from the resulting .wasm file. 
>
> When prototyping or testing locally, you might want to avoid `--release` to obtain useful backtraces in case of errors (for example, with `wasmtime::WasmBacktraceDetails::Enable`). Note: the resulting .wasm file will be considerably larger (likely 4MB+).
```

vados-cosmonic (approved):
LGTM :rocket: 

One nit about an alert but this looks good, let's get this merged :)

component-model/examples/tutorial/adder/src/lib.rs
vados-cosmonic: Yeah I think this is fine! Note that the `docs:adder` is `namespace:package`, so dunno if that's worth separating out. But I am definitely all for the comments :)

mkatychev:
@kate-goldenring this PR should be ready for a final review & merge

component-model/src/language-support/rust.md
```diff
@@ -1,101 +1,103 @@
...
+## 2. Scaffolding a Component
 
-## 3. Add the WIT world the Component will implement
+We will create a component in Rust that implements the `add` function exported
+by the [`adder` world][docs-adder] world in the
+`docs:adder`
+[package](https://github.com/WebAssembly/component-model/blob/main/design/mvp/WIT.md#package-names).
 
-Next, update `wit/world.wit` to match the [`adder` world][adder-wit]:
+First `cd` into the `tutorial` directory found in the repo we just cloned:
```
kate-goldenring, line 37: Why do you need to be in this directory? Can we just have them fetch the wit instead? Once we have the wits published we can go a `wkg get` here. For now, so they don't have to clone the repo the following works:
`wget https://raw.githubusercontent.com/bytecodealliance/component-docs/refs/heads/main/component-model/examples/tutorial/wit/adder/world.wit`

component-model/src/language-support/rust.md
```diff
@@ -1,101 +1,103 @@
...
-
-bindings::export!(Component with_types_in bindings);
+{{#include ../../examples/tutorial/adder/src/lib.rs}}
 ```
 
-## 6. Build the component
+## 6. Building a Component
 
-Now, use `cargo component` to build the component, being sure to optimize with a release build.
+Now, let's build our component, being sure to optimize with a release build:
```
kate-goldenring, line 86: Just curious, what motivated changing from the previous wording?

component-model/src/language-support/rust.md
```diff
@@ -1,101 +1,103 @@
...
 
 ```console
 cargo component build --release
 ```
 
-> **WARNING:** Building with `--release` removes all debug-related information from the resulting .wasm file. When prototyping or testing locally, you might want to avoid `--release` to obtain useful backtraces in case of errors (for example, with `wasmtime::WasmBacktraceDetails::Enable`). Note: the resulting .wasm file will be considerably larger (likely 4MB+).
-
-You can use `wasm-tools component wit` to output the WIT package of the component:
+You can use `wasm-tools` to output the WIT package of the component:
```
kate-goldenring, line 92: Please bring back these warnings. This may have been a mistake during a rebase, but they were recently added by a contributor: https://github.com/bytecodealliance/component-docs/pull/234

component-model/src/language-support/rust.md
```diff
@@ -1,101 +1,103 @@
...
 
+```console
```
kate-goldenring, line 94: Please revert this to show the output in the console block. See this issue for reference https://github.com/bytecodealliance/component-docs/issues/239

component-model/src/language-support/rust.md
```diff
@@ -108,10 +110,15 @@ package docs:adder@0.1.0 {
 }
 ```
 
-### Running a Component from Rust Applications
+> [!WARNING]
```
kate-goldenring, line 113: Please remove this and move this chunk back to where it previously was after the `cargo component build --release` command

component-model/src/language-support/rust.md
```diff
@@ -150,14 +175,19 @@ world calculator {
 
 ### Referencing the package to import
 
-Because the `docs:adder` package is in a different project, we must first tell `cargo component` how to find it. To do this, add the following to the `Cargo.toml` file:
+Because the `docs:adder` package is in a different project, we must first tell
```
kate-goldenring, line 178: Are there any changes here?

component-model/src/language-support/rust.md
```diff
@@ -150,14 +175,19 @@ world calculator {
...
+`cargo component` how to find it. To do this, add the following to the
+`Cargo.toml` file:
 
 ```toml
 [package.metadata.component.target.dependencies]
-"docs:adder" = { path = "../adder/wit" }  # directory containing the WIT package
+"docs:adder" = { path = "../wit/adder" }  # directory containing the WIT package
```
kate-goldenring, line 184: This assumes they cloned the repo rather than just fetching the wit. I'd prefer to stay with the approach of fetching the one wit file rather than cloning the repo

component-model/src/language-support/rust.md
```diff
@@ -262,8 +292,8 @@ As mentioned above, `cargo component build` doesn't generate a WIT file for a co
 
     ```toml
     [package.metadata.component.target.dependencies]
-    "docs:calculator" = { path = "../calculator/wit" }
-    "docs:adder" = { path = "../adder/wit" }
+    "docs:calculator" = { path = "../wit/calculator" }
```
kate-goldenring, line 295: Same comment as above

kate-goldenring (changes requested):
Thank you @mkatychev for these updates. My main suggestion is to not require the reader to clone the repository. Rather, have them just fetch the WIT file.

component-model/src/language-support/rust.md
mkatychev: The tutorial steps involve interactive with the adder package [here](https://component-model.bytecodealliance.org/language-support/rust.html#running-a-component-from-rust-applications) and the 3rd step [here ](https://component-model.bytecodealliance.org/language-support/rust.html#importing-an-interface-into-a-command-component)references multiple directories found in the the component docs repo.

component-model/src/language-support/rust.md
mkatychev: Following up on https://github.com/bytecodealliance/component-docs/pull/218#discussion_r2037888048, if we don't specify cloning the repo then we'd have to specify fetching the `wit/adder`, `wit/calculator`, and `examples/adder` directories explicitly, I don't mind doing this but it seems like that adds more complexity than a `git clone`.

`git clone` is a temporary approach regardless until https://github.com/bytecodealliance/component-docs/issues/241#issuecomment-2794528773 is done.

kate-goldenring: I think fetching a file is the most analogous to what the wkg experience will be like

component-model/src/language-support/rust.md
mkatychev: Yeah totally my mistake.

component-model/src/language-support/rust.md
kate-goldenring: Sounds good. We can leave this change. Going forward, in order to scope PRs, we may want to avoid changing tense until we've established a pattern in `CONTRIBUTING.md`.

component-model/src/language-support/rust.md
mkatychev: Would you mind elaborating? 
```
"docs:calculator" = { path = "../calculator/wit" }
"docs:adder" = { path = "../adder/wit" }
https://component-model.bytecodealliance.org/language-support/rust.html#running-a-component-from-rust-applications

component-model/src/language-support/rust.md
kate-goldenring: I agree that the example host is an exception -- for now, they will need to clone the repo to run it. Maybe this line should be `cargo run --release -- 1 2 adder.wasm` to not imply that they build the adder component from the repo. 

The main thing i am thinking about is that the fact that the example host is not fetchable should not define how we instruct users to build components. We can push the host up in a container for example or release it.

kate-goldenring: Is the example host confusing? Should we not provide an option to execute the add function at all? Alternatively, we could provide a command component that imports the add interface that could be composed with it then that final component could be executed with wasmtime

component-model/src/language-support/rust.md
mkatychev: Would you mind elaborating on where/when the reader should have these directories present?

The docs reference three external directories as part of the Rust section (excluding the example-host mentioned in https://github.com/bytecodealliance/component-docs/pull/218#discussion_r2037987446):
* [`"docs:calculator" = { path = "../calculator/wit" }` and `"docs:adder" = { path = "../adder/wit" }`](https://component-model.bytecodealliance.org/language-support/rust.html#running-a-component-from-rust-applications)
* [`"docs:rpn" = { path = "../wit" }`](https://component-model.bytecodealliance.org/language-support/rust.html#running-a-component-from-rust-applications)

Should we add explicit steps to create and populate those directories (or I might have missed those steps)?

kate-goldenring: In my opinion, the reader should put the wits wherever they want. Maybe that means this should be `"docs:adder" = { path = "../path/to/wit" }`. To me, this is a language guide that is explaining concepts through an example.

component-model/src/language-support/rust.md
macovedj: Does it make sense to have a wit in ghcr that they can depend on?  Then there would be no question about file paths because they wouldn't be needed when specifying the dependency.

component-model/src/language-support/rust.md
mkatychev: Should we remove the explicit relative filepaths like `../calculator/wit`, change them to docstring style paths like `../path/to/your/calculator/wit`, or remove the toml codeblocks entierly?

I'm happy to go with an implicit approach, pruning/changing the explicit references would then alleviate some of the confusion.

component-model/src/language-support/rust.md
kate-goldenring: Created an issue to get them in GHCR: https://github.com/bytecodealliance/component-docs/issues/243

component-model/src/language-support/rust.md
mkatychev: No, I think this was retained as a reflow of changes I later reverted. [Some markdown styleguides](https://google.github.io/styleguide/docguide/style.html#character-line-limit) suggest keeping markdown lines to an 80 char soft limit where you can to help with editability.

component-model/src/language-support/rust.md
mkatychev: In the meantime for this PR, should I revert the path changes and references to `git clone`?

kate-goldenring (approved):
Thank you for making this a smoother experience!

PR #222, opened 2025-03-24
fix(lang/rust): link to old WIT, prose & interface use example

vados-cosmonic:


component-model/src/language-support/rust.md
```diff
@@ -122,37 +122,71 @@ $ cargo run --release -- 1 2 ../add/target/wasm32-wasip1/release/add.wasm
 
 ## Exporting an interface with `cargo component`
 
-The [sample `add.wit` file](https://github.com/bytecodealliance/component-docs/tree/main/component-model/examples/example-host/add.wit) exports a function. However, you'll often prefer to export an interface, either to comply with an existing specification or to capture a set of functions and types that tend to go together. For example, to implement the following world:
+While it is possible to export a raw function, prefer exporting an interface, similar to the [sample `adder` world][add-wit].
```
kate-goldenring, line 125:
```suggestion
While it is possible to export a raw function, it is a best practice to export an interface, similar to the [sample `adder` world][add-wit]. 
```

component-model/src/language-support/rust.md
```diff
@@ -122,37 +122,71 @@ $ cargo run --release -- 1 2 ../add/target/wasm32-wasip1/release/add.wasm
...
+        Operand::Unsigned32(v) => v as i64,
+        Operand::Signed32(v) => v as i64,
+        Operand::Zero => 0,
     }
 }
+
+bindings::export!(Component with_types_in bindings);
 ```
 
+While the code above is unlikely to appear in real interfaces, but it shows the more common case of an interface
```
kate-goldenring, line 187:
```suggestion
The code above is unlikely to appear in real interfaces; however, it shows the more common case of an interface 
```

component-model/src/language-support/rust.md
kate-goldenring, line 125: I am not sure I follow why you made these changes to the "Exporting an interface section". Is there a reason you wanted to add in the concept of exporting types too? I may make sense to clarify that to the reader. Maybe: In this section we will export a function and the types it uses to be consumed by another component. 

The way it is written currently, it seems like adding the operand type is related to the general concept of exporting an interface

vados-cosmonic: Ah, so my thinking here was that the above code is *already* exporting the interface, so there was no reason to call that out specifically in a section.

I originally deleted this section, then thought it would be a waste of a good spot to put an example, so I leaned in to the note about "capture a set of functions and types that tend to go together", and wrote an example about that.

> The way it is written currently, it seems like adding the operand type is related to the general concept of exporting an interface

Did you mean "unrelated" here?

component-model/src/language-support/rust.md
```diff
@@ -122,37 +122,71 @@ $ cargo run --release -- 1 2 ../add/target/wasm32-wasip1/release/add.wasm
...
+
+For example, consider the following changes to the `add` interface in the `adder` world:
 
 ```wit
 package docs:adder@0.1.0;
 
+
 interface add {
-    add: func(x: u32, y: u32) -> u32;
+    variant operand {
```
itowlson, line 136: For myself, I'd prefer to see this introduced in a more granular, step-by-step manner.  This section is showing how to move from exporting a loose function to exporting an interface.  Let's try to keep the rest of it the same so the reader can easily see "ah this is what has changed in the WIT and this is how it changes the Rust."  To me, the new stuff about operand conversion gets in the way of that.

I am totally up for showing types in interfaces and how they surface in Rust but for myself I'd be inclined to split that across two stages.  "Now you've seen an interface, let's make the interface FANCY-SCHMANCY."

component-model/src/language-support/rust.md
```diff
@@ -17,19 +17,19 @@ cargo install cargo-component
...
 ## 2. Scaffold a Component with `cargo component`
 
-Create a Rust library that implements the `add` function in the [`adder`world][adder-wit].
+Create a Rust library that implements the `add` function in the [`adder` world][adder-wit].
 
 First scaffold a project:
 
 ```sh
-$ cargo component new add --lib && cd add
+$ cargo component new adder --lib && cd adder
```
itowlson, line 25: I believe if you change this you'll need to change lines 96 and 119, as the .wasm file name will change?

component-model/src/language-support/rust.md
t3hmrman: Thanks for the feedback -- I guess for now, I'll remove these changes, it is clearly too much to introduce at once!

component-model/src/language-support/rust.md
t3hmrman: Done! thanks for the note -- updated the wasm file names

component-model/src/language-support/rust.md
t3hmrman: OK, so this section got removed (see [discussion here](https://github.com/bytecodealliance/component-docs/pull/222#discussion_r2015010725)), should be fine now!

t3hmrman:
Hey @itowlson @kate-goldenring thanks for the reviews -- this should be ready for another look!

component-model/src/language-support/rust.md
vados-cosmonic: OK, so this section got removed (see https://github.com/bytecodealliance/component-docs/pull/222#discussion_r2015010725), should be fine now!

component-model/src/language-support/rust.md
vados-cosmonic: Done! thanks for the note -- updated the wasm file name!

vados-cosmonic:
Hey itowlson kate-goldenring thanks for the reviews -- this should be ready for another look!

component-model/src/language-support/rust.md
itowlson: @vados-cosmonic From the diff it looks like you've deleted the section rather than just reverting the changes?  Sorry for the miscommunication, that's not what I was trying to suggest.  Or am I misreading?

vados-cosmonic: Ah yes, that's right -- that section actually doesn't make sense anymore, because we *always* export an interface (rather than a function directly)

itowlson: Oh!  I completely missed that!  Yes, then there is definitely no need for it.  (And if I'd realised the page had already done interfaces, I wouldn't have been bothered by adding the operand stuff... the section title implied this was the first introduction of "how to export an interface" and I didn't double check.  I'm sorry about that!)

itowlson (approved):
Many apologies for my confusion over this, and thanks for your patience @vados-cosmonic!

vados-cosmonic:
Absolutely no problem! Yeah the example was doing the other thing until just this week/last week so I totally get it. Thanks for taking another look!

PR #223, opened 2025-03-24
Update python.md

ingydotnet:
Remove extra space in Python code that would make it print:
```
1 + 2 =  3
```
instead of:
```
1 + 2 = 3
```


itowlson (approved):
Thanks!

PR #228, opened 2025-03-25
fix(ci): bash check

vados-cosmonic:


itowlson (approved):

vados-cosmonic:
Hey @itowlson let's roll back -- the python sitemap generation step is going to fail because the `scripts` path needs to be one step up to get "back" as the script is currently written. 

Thanks for your help, will prep a PR that goes back to the python PR that merged

vados-cosmonic:
OK so unfortunately this is even worse because the merge commits make it a bit hard to revert *and* there are a LOT of commits to go backwards on..

PR #229, opened 2025-03-25
fix(ci): sitemap generation

vados-cosmonic:


vados-cosmonic:
Hey @itowlson so I think we have to go back to here https://github.com/bytecodealliance/component-docs/commit/24d7eb9e063074a01bc1393bb377165a1ce6df40, which is a bit difficult to do with the merge commits in the history. I'm trying to fix the python step instead -- it relies on relative paths so the output directory change broke that too unfortunately.

vados-cosmonic:
If you wouldn't mind just *one* more time :crossed_fingers: , I think this will finally fix all the steps for deploy.

itowlson:
Thanks and belatedly merged!

vados-cosmonic:
Thanks for the patience here, glad to note that [the docs are back](https://component-model.bytecodealliance.org/) :sweat_smile: 

It's unfortunate it took this many steps to iron out CI -- hopefully I can make this process even more error-proof as part of https://github.com/bytecodealliance/component-docs/pull/226 . I think it might be useful to:

- Change (and maybe factor out) the build process to produce a zip file artifact
- Introduce a top level script file (I am partial to [`just`](https://github.com/casey/just) as an ergonomic `make` replacement) so that we stop `cd`ing around in CI
- Update the python script to only deal with absolute/script-relative paths (running via `just` obviates this but this is a nice to have)
- Refactor publish to only take the artifact and then deploy to pages
- Enable publish workflow to work on *any* commit (this way, we can "roll back" without actually rolling back code), so we could have just rolled back to a working commit the first time CI went wrong :face_exhaling:

itowlson:
Thank you for working through the series of fixes!  Your suggestions for making this more reliable and understandable sound good - in particular I'm 100% up for running a lot more of this stuff during the PR process.  I also like the idea of a manual publish workflow for rapid rollback - that speaks to my risk-averse soul!

Issue #231, opened 2025-03-25
Move relevant `sh` codeblocks to `console` 


mkatychev:
> yeah so my stance on that is that writing into a console is not normally syntax highlighted either, so it matches.
>
> If we have a `sh`ell script, or a `bash` script, then it makes sense to use those, but my stance here is that if it's a console command you're supposed to enter there are other ways to make it more readable:
>
> ```
> something \
>     --like this \
>     --rather-than console
> ```
>
> and if it's code, *then* it should be highlighted.

_Originally posted by @vados-cosmonic in https://github.com/bytecodealliance/component-docs/pull/218#discussion_r2011220029_

itowlson:
It would be useful to have an example of what difference this would make to the rendered output.  (I'm not sure if I need to address this comment to Victor or Mikhail though, sorry!)

mkatychev:
This should be sent to @vados-cosmonic, I just split his https://github.com/bytecodealliance/component-docs/pull/218#discussion_r2011220029 into a separate PR

vados-cosmonic:
Hey @mkatychev thanks for filing this issue -- to recap context for @itowlson (please correct me if I'm wrong).

- Mikhail noted that we would lose syntax highlighting for a block that was `sh`
- I noted that things that are entered into the console are usually *not* highlighted to begin with, and that `sh`/`bash`/etc where more for scripts than for things you enter at the console (the quote above), IMO

I think Mikhail took this as a good chance to go around the codebase and make the change with that philosophy in mind (thanks!), but I asked for a issue that we could discuss this, since I think it could be a point of contention. 

I'm happy to change my stance on this, but to sum up:

- `console` for things people should paste in
- `sh`/`bash`/etc for actual scripts (that people should copy into a file and run)
- `<lang>` for code snippets

I personally like the style of demarcating commands and the output people should see in prose, but that's somewhat separate!

mkatychev:
@vados-cosmonic if memory serves right, an issue was only suggested for `mdbook-alerts` (which has been made). I'm happy to put this in draft and link an issue for this PR.

vados-cosmonic:
Sorry I think I should have been clearer -- I was thinking that *before* we go make the whole codebase one way or another, we should probably get input/discuss it first! 

I think once people have put in some input we can figure out how to proceed? Then we'll know for the future/can avoid having to actively make the decision again if someone else contributes.

itowlson:
[...]

PR #232, opened 2025-03-25
chore(docs): Move relevant `sh` codeblocks to `console` 

mkatychev:
Move certain codeblocks that are not `sh` code to `console`, issue: https://github.com/bytecodealliance/component-docs/issues/231

### DIFF
--- a/component-model/examples/tutorial/README.md
+++ b/component-model/examples/tutorial/README.md
@@ -54,7 +54,7 @@ wac plug command/target/wasm32-wasip1/release/command.wasm --plug composed.wasm 
 
 Now, run the component with Wasmtime:
 
-```sh
+```console
 wasmtime run final.wasm 1 2 add
 1 + 2 = 3
 ```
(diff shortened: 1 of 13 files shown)


component-model/examples/tutorial/README.md
```diff
@@ -101,7 +101,7 @@ export command-instance["wasi:cli/run@0.2.0"];
 Now, perform your composition by passing the WAC file to `wac compose`.
 
 ```sh
-wac compose composition.wac -o final.wasm 
+wac compose composition.wac -o final.wasm
```
vados-cosmonic, line 104: I think this could be `console` if we're doing that!

component-model/src/creating-and-consuming/composing.md
```diff
@@ -94,7 +94,7 @@ export validator...;
 
 Then, `wac compose` can be used to compose the components, passing in the paths to the components. Alternatively, you can place the components in a `deps` directory with an expected structure, and in the near future, you will be able to pull in components from registries. See the [`wac` documentation](https://github.com/bytecodealliance/wac) for more details.
 
-```sh
+```console
```
vados-cosmonic, line 97: This would be a good time to space out the command line args to different lines so the block is a bit easier to read.

component-model/src/language-support.md
```diff
@@ -79,7 +79,7 @@ working with WebAssembly modules and components.
 4. Use `wasm-tools` to create a component from the core module, first embedding component metadata
    inside the core module and then encoding the WAT to a Wasm binary.
 
-   ```sh
+   ```console
```
vados-cosmonic, line 82: If we have `console` and only commands in a block, we can take out the `$`, because it makes it hard to copy paste

Ditto for the other places in the PR as well!

component-model/src/language-support/csharp.md
```diff
@@ -199,7 +199,7 @@ public class AddImpl : IAdd
 
 Once again, compile an application to a Wasm component using `dotnet build`:
 
-```sh
+```console
```
vados-cosmonic, line 202: No need to do anything here, but this actually *shouldn't* be a `console` block IMO, because it contains the command *and* output.

I would normally leave this *without* a highlighting tag all together, but that's me.

@itowlson this is a good example of the nuance IMO -- would love to know your thoughts here.

component-model/src/language-support/csharp.md
```diff
@@ -275,7 +275,7 @@ Console.WriteLine($"{left} + {right} = {result}");
 
 Once again, compile your component with `dotnet build`:
 
-```sh
+```console
```
vados-cosmonic, line 278: Same note here as above w/ the combination of command and output

mkatychev:
PR had been closed pending further discussion on using `console` language for codeblocks

PR #234, opened 2025-04-02
Add documentation warnings for Rust, Go, and JavaScript

alexcasalboni:
3 main changes:

1. Add a `--release` warning in the Rust docs to clarify that always building with `--release` might not be ideal if you're prototyping/testing locally and you need debug info in your backtraces
2. Add a `-no-debug` warning in the Go docs to clarify that by default tinygo includes all debug information and that you can optimize your .wasm file by using `-no-debug` (up to 75% reduction in size)
3. Add a `--disable all` warning in the JS docs since many users might actually need to use `console.log` or `new Date()` in their component for logging/debugging


kate-goldenring (approved):
Thank you for adding these clarifiers!

PR #244, opened 2025-04-10
chore(infra): Add `.gitattributes` file to repo

mkatychev:
Added a `.gitattributes` file to help [github-linguist understand](https://github.com/github-linguist/linguist/blob/main/docs/overrides.md) what files are binaries or generated code.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,8 @@
+# Ensure all text is checked out with Unix EOL
+* text=auto eol=lf
+
+# Ensure binary files aren't considered as text
+*.wasm binary
+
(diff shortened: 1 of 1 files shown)


vados-cosmonic (approved):
LGTM :rocket: 

I'm assuming that this won't add a dependency on linguist for everyone that pulls right? So that's only for GitHub to pick up?

mkatychev:
That is correct, this is just a way to demarcate generated code & binaries

vados-cosmonic:
Thanks for clarifying! TIL of [linguist](https://github.com/github-linguist/linguist/tree/main) :)

PR #260, opened 2025-05-15
feat: add section with video links

### LINKED ISSUE #127: Link to component model related talks
Some people prefer to learn by watching. @rylev gave a great overview of the component model in her talk at Wasm I/O:  https://youtu.be/zqfF7Ssa2QI?si=3JV_zvzc1XZZFQhi

vados-cosmonic:
Resolves #127

### DIFF
--- a/component-model/src/reference/videos.md
+++ b/component-model/src/reference/videos.md
@@ -1,0 +1,34 @@
+# Videos
+
+The WebAssembly ecosystem often gives talks about functionality and features of WebAssembly components.
+
+This page is a non-exhaustive list of some of the best talks and videos related to the component model.
+
+## Overviews of the Component Model
+
+[Threading the needle with concurrency and parallelism in the Component Model - Luke Wagner](https://www.youtube.com/watch?v=mkkYNw8gTQg)
+
+[The Path to Components - Luke Wagner](https://www.youtube.com/watch?v=phodPLY8zNE)
+
+[What is a Component (and Why)? - Luke Wagner](https://www.youtube.com/watch?v=tAACYA1Mwv4)
+
+[Deconstructing WebAssembly Components - Ryan Levick](https://www.youtube.com/watch?v=zqfF7Ssa2QI)
+
+## Full Conference playlists
+
+> [!WARNING]
+> Exercise caution when watching conference talks and videos older than ~2024.
+>
+> WebAssembly specifications have changed quite a bit since 2023 and proposed versus implemented
+> semantics (Component Model, WIT, etc) have seen considerable improvements.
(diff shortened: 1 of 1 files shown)


itowlson (approved):
This is a great starting point - thanks!

I would maybe suggest some caution around older talks.  There has been a lot of churn even since 2023 and I'd be cautious about encouraging people to look to those talks for current information.  Of course this will be an ongoing issue for all our outbound links though!

vados-cosmonic:
Ah great point, I'll do a bit of pruning before I move this forward!

vados-cosmonic:
OK So it seemed a bit fraught to actually cut out WASM I/O 2023 (the first year!) so what I opted for was a a warning in the conference playlist section -- what do you think @itowlson ?

itowlson:
@vados-cosmonic The basic warning is great, but I'd drop the specific example (and definitely wouldn't link to WITX).  Focus the reader on the general problem of churn and obsolescence, rather than risking them zeroing in on one change and being less cautious about other (equally obsolete) content.

I do worry that this is going to be a broad issue with full conference playlists, though, and wonder if we should establish a principle like "only full lists from conferences in the last 2 years, otherwise curate."  As long as the conference name is mentioned people can search up older ones.  But that's something we can ponder over time and we don't need to tackle it for this PR.

vados-cosmonic:
Done! Thanks for the input @itowlson! :rocket: 

For now we'll make an exception for WASM I/O (since it's kind of... our flagship conference?)

component-model/src/reference/videos.md
```diff
@@ -0,0 +1,34 @@
...
+
+### Wasm I/O
+
+[Wasm I/O 2025 Playlist](https://www.youtube.com/playlist?list=PLP3xGl7Eb-4OtFH1tcQm6u7_LRED7-3rg)
+
+[Wasm I/O 2024 Playlist](https://www.youtube.com/playlist?list=PLP3xGl7Eb-4Nmj4CJ5WLQZx5UAYvhH920)
+
+[Wasm I/O 2023 Playlist](https://www.youtube.com/playlist?list=PLP3xGl7Eb-4NtSX-wzOxUCjPsBcw2zrPA)
+
+[witx]: https://github.com/WebAssembly/WASI/tree/main/legacy/preview1/witx
```
itowlson, line 33: Might want to prune these now unused links - otherwise LGTM - thanks!

vados-cosmonic: Ah forgot to remove these, made another PR:

https://github.com/bytecodealliance/component-docs/pull/261

PR #267, opened 2025-06-17
remove 6.1 'Authoring Components' section

macovedj:
Closes https://github.com/bytecodealliance/component-docs/issues/263.  For context, in the SIG Docs call, there was discussion around reorganizing chapter 6.  One of the take aways was that this section does not add much information to the chapter as a whole.


itowlson:
In a section titled "creating and consuming components," one of my first questions is going to be "so, creating?"  So I do think it's worth having _something_ whose title clearly expresses "start here if you want to know how to create a component."  I agree the current page doesn't add much, but having some kind of signpost at the TOC level feels to me like it has some value.

Issue #264, opened 2025-05-21
Remove section 6.3


kate-goldenring:
The content is out of date and best fit in the Wasmtime section

vados-cosmonic:
Resovled by #277

PR #268, opened 2025-06-17
running.md moved to wasmtime.md section and updated w custom invocations

macovedj:
closes https://github.com/bytecodealliance/component-docs/issues/264, also as part of updates to chapter 6 discussed on SIG Docs meeting.


component-model/src/creating-and-consuming/running.md
itowlson, line 1: Does mdbook give us a way to keep a redirect here, just in case?

Also, I again have a bit of a concern re signposting.  The "creating and consuming" section provides a kind of journey of I write my component and I compose it with other components and then I run it and then I publish it.  And removing this leaves a gap.  The reader now needs to know that "how do I run my component" lives in the "Runtime Support" section, the title of which feels like it's going to be a list of caveats and feature comparisons for different runtimes rather than "this is how you make your component go."

Issue #262, opened 2025-05-21
document `wasmtime run --invoke`


kate-goldenring:
Blog https://bytecodealliance.org/articles/invoking-component-functions-in-wasmtime-cli

PR #277, opened 2025-07-10
Refactor 6

macovedj:
Refactors the sixth chapter into one chapter on running components, and another on composition and distribution of components.

Combines https://github.com/bytecodealliance/component-docs/pull/268 and https://github.com/bytecodealliance/component-docs/pull/267.

It also closes https://github.com/bytecodealliance/component-docs/issues/262, https://github.com/bytecodealliance/component-docs/issues/264, and https://github.com/bytecodealliance/component-docs/issues/263.


component-model/book.toml
```diff
@@ -9,6 +9,14 @@ title = "The WebAssembly Component Model"
 git-repository-url = "https://github.com/bytecodealliance/component-docs/tree/main/component-model"
 edit-url-template = "https://github.com/bytecodealliance/component-docs/tree/main/component-model/{path}"
 additional-css = ["theme/head.hbs"]
+[output.html.redirect]
```
vados-cosmonic, line 13:
```suggestion

[output.html.redirect]
```

NIT: I do like me some space between these sections

Thanks for adding this... I think we've just been dropping links after renames up until now.

What do you think about how to handle this over time? Do we keep this list of redirects going forever? If so, it might be awesome if we could make this a separate file.

component-model/src/SUMMARY.md
```diff
@@ -22,18 +22,14 @@
   - [JavaScript](./language-support/javascript.md)
   - [Python](./language-support/python.md)
   - [Rust](./language-support/rust.md)
-- [Creating and Consuming Components](./creating-and-consuming.md)
-  - [Authoring Components](./creating-and-consuming/authoring.md)
-  - [Composing Components](./creating-and-consuming/composing.md)
-  - [Running Components](./creating-and-consuming/running.md)
-  - [Distributing and Fetching Components and WIT](./creating-and-consuming/distributing.md)
+- [Running Components](./running-components.md)
```
vados-cosmonic, line 25: Since we've renamed this section, what about renaming the "Language Support" section to be more like "Creating components" ? Would fit really well with this new organization

component-model/src/introduction.md
```diff
@@ -5,8 +5,8 @@ The WebAssembly Component Model is a broad-reaching architecture for building in
 | Understanding components | Building components  | Using components  |
 |--------------------------|----------------------|-------------------|
 | [Why Components?]        | [C/C++]              | [Composing]       |
-| [Components]             | [C#]                 | [Running]         |
-| [Interfaces]             | [Go]                 | [Distributing]    |
+| [Components]             | [C#]                 | [Distributing]    |
```
vados-cosmonic: So IMO Running was valuable here... IMO if you build a component the first thing you do is run it, not compose it, but maybe that's just a personal preference.

The "Running" link did get updated at the bottom, did it maybe just get left out?

component-model/src/running-components/wasmtime.md
```diff
@@ -1,6 +1,6 @@
 # Wasmtime
 
-[Wasmtime](https://github.com/bytecodealliance/wasmtime/) is the reference implementation of the Component Model. It supports running components that implement the [`wasi:cli/command` world](https://github.com/WebAssembly/wasi-cli/blob/main/wit/command.wit) and serving components that implement the [`wasi:http/proxy` world](https://github.com/WebAssembly/wasi-http/blob/main/wit/proxy.wit).
+[Wasmtime](https://github.com/bytecodealliance/wasmtime/) is the reference implementation of the Component Model. It supports running components that implement the [`wasi:cli/command` world](https://github.com/WebAssembly/wasi-cli/blob/main/wit/command.wit) and serving components that implement the [`wasi:http/proxy` world](https://github.com/WebAssembly/wasi-http/blob/main/wit/proxy.wit). It can also invoke functions exported from a component.
```
vados-cosmonic: What do you think about replacing "It" with "Wasmtime" in this paragraph? Feels like the clarity benefit would outweigh the repetition.

Maybe we could even make this a relatively short bulleted list

component-model/src/running-components/wasmtime.md
```diff
@@ -27,3 +27,13 @@ Try out building and running HTTP components with one of these tutorials
 1. [Hello WASI HTTP tutorial](https://github.com/sunfishcode/hello-wasi-http) - build and serve a simple Rust-based HTTP component
 
 2. [HTTP Auth Middleware tutorial](https://github.com/fermyon/http-auth-middleware#running-with-wasmtime) - compose a HTTP authentication middleware component with a business logic component
+
+## Running components with custom exports
+As of Wasmtime Version 33.0.0, there is [support for invoking custom component exports](https://bytecodealliance.org/articles/invoking-component-functions-in-wasmtime-cli).
```
vados-cosmonic:
```suggestion
As of Wasmtime Version 33.0.0, there is [support for invoking components with custom exports](https://bytecodealliance.org/articles/invoking-component-functions-in-wasmtime-cli).
```

I'm not tied to this change but what do you think? Does it increase clarity of what the new `--invoke` can do?

component-model/src/running-components/wasmtime.md
```diff
@@ -27,3 +27,13 @@ Try out building and running HTTP components with one of these tutorials
...
+
+
+As an example, if your component exports a function `add` which takes two numeric arguments, you can make use of this feature with the following command.
```
vados-cosmonic, line 35: Does it make sense to explain a little bit about what the world was like *before* this new feature? Like a small note about how there wasn't a general way to pass in argument types, even when they were simple values.

Not tied to this, IMO what's here is fine.

Also what do you think about `component.wasm` and using phrasing like "Given a component `adder.wasm`" or something like that? We'd be able to make that wasmtime run command below a bit more "concrete".

component-model/src/running-components/wasmtime.md
```diff
@@ -27,3 +27,13 @@ Try out building and running HTTP components with one of these tutorials
...
+
+`wasmtime run --invoke 'foo(1, 2)' <path-to-wasm-file>`
```
vados-cosmonic: This could be code block?

vados-cosmonic (approved):
LGTM :rocket: 

I think this could land as-is, but I left some comments (with some nits) that are worth discussing just a tiny bit

component-model/src/SUMMARY.md
macovedj: I actually considered this myself, but I think the fact that it's an enumeration of how to author components in each language, and there are plenty of languages that aren't included, I ended up leaning towards keeping it as is.  Would be nice if there were a way to get both without making the chapter name clunky.

component-model/src/running-components/wasmtime.md
macovedj: How about `invoking components with custom exports`?  I feel like the exports themselves are user-defined, and the "invoker" can provide inputs dynamically.  But idk I don't feel too strongly one way or the other.

component-model/src/running-components/wasmtime.md
macovedj: Hmm.  Personally I used to bump into this all the time and had to do a jco transpile whenever I had this use case,  and this is definitely a much smoother than that.  I see the value of adding some context here, but I'm not sure how to do so without just giving the reader more to wade through, rather than concise and clear instructions.  If you have something in mind I'm happy to take a contribution.

component-model/src/SUMMARY.md
t3hmrman: Ahh, could you explain this part:

> it's an enumeration of how to author components in each language, and there are plenty of languages that aren't included

If it's an enumeration of how to author/create components, then maybe we should name the section as such? "Language support" kind of evokes a "for what?" and if the answer is "Component Model" that's probably quite a different section (i.e. it would be more about P1 vs P2, whether you need to adapt, etc)

But actually looking at [what we have now](https://component-model.bytecodealliance.org/language-support.html), this isn't the only problem with that page -- there is a section that is for some reason grouped in with the header that should probably be it's own page.

Maybe this is worth pushing to a follow up

component-model/src/running-components/wasmtime.md
t3hmrman: Oh yeah that sounds great -- I'm happy either way but "invoking components with custom exports" sounds like an improvement.

The word "custom" carries some weight there (I think the implication is "as opposed to standardized wasi exports"), but impossible to get that in here.

component-model/src/SUMMARY.md
vados-cosmonic: Ahh, could you explain this part:

> it's an enumeration of how to author components in each language, and there are plenty of languages that aren't included

If it's an enumeration of how to author/create components, then maybe we should name the section as such? "Language support" kind of evokes a "for what?" and if the answer is "Component Model" that's probably quite a different section (i.e. it would be more about P1 vs P2, whether you need to adapt, etc)

But actually looking at [what we have now](https://component-model.bytecodealliance.org/language-support.html), this isn't the only problem with that page -- there is a section that is for some reason grouped in with the header that should probably be it's own page.

Maybe this is worth pushing to a follow up

component-model/src/running-components/wasmtime.md
vados-cosmonic: Oh yeah that sounds great -- I'm happy either way but "invoking components with custom exports" sounds like an improvement.

The word "custom" carries some weight there (I think the implication is "as opposed to standardized wasi exports"), but impossible to get that in here.

component-model/src/running-components/wasmtime.md
vados-cosmonic: Yeah this is true -- it would definitely be a lot to try and take in, too much potentially.

Issue #122, opened 2024-03-25
Remove most mentions of the "component model"


rylev:
For almost all of usages of the term "component model", we would likely be better off just saying "WebAssembly components". Usage of the term "component model" naturally leads to the question of what is the "model" in "component model" which it self requires explaining what components are. By moving to "WebAssembly components" we cut straight to the relevant question, the reader is likely to be asking, namely: what are WebAssembly components?

kate-goldenring:
Great point. +1

vados-cosmonic:
Hey @catamorphism do you think we can close this now? I think the work you just landed mostly solves this right?

sahidulsany:
Facebook code

mkatychev:
We define the term [component model](https://component-model.bytecodealliance.org/introduction.html#:~:text=The%20WebAssembly%20Component%20Model%20is%20a%20broad%2Dreaching%20architecture%20for%20building%20interoperable%20WebAssembly%20libraries%2C%20applications%2C%20and%20environments%2E) early on so most readers have read the introduction should be aware of the term. Feel free to reopen if this warrants further discussion.

PR #280, opened 2025-07-15
Revise introduction

catamorphism:
By way of introduction, I'm Tim Chevalier at Igalia; we're being sponsored to work on improving the component model documentation, especially with an eye towards making it more understandable by WebAssembly beginners. 

This PR reorganizes the introduction by defining WASI before it's mentioned, and making a few other changes to make the text flow better. I also changed "Component Model" to "WebAssembly Components" in the sidebar per #122. Finally, I fixed a broken link in the FAQ section.

I attempted to use semantic line breaks in my changes, but if there's a preferred style for these docs that precludes that, let me know.

### DIFF
--- a/component-model/src/introduction.md
+++ b/component-model/src/introduction.md
@@ -37,7 +37,38 @@ This documentation is aimed at _users_ of the component model: developers of lib
 [Distributing]: ./composing-and-distributing/distributing.md
 
 
+## WebAssembly components
+
+As with all programming, the goal of writing a component
+is to make new functionality available
+by building it out of existing functionality.
+
+A WebAssembly component runs on a _platform_,
+which may be a Web browser,
+a stand-alone runtime,
+or even an operating system (when compiling WebAssembly to an executable).
+By running the component, the platform gains the functionality
+that the component implements.
+Likewise, the platform provides functionality
+that code in components can use to interact
+with the outside world.
+
+For example:
+
+- A user of the component model can build a component
+  that converts the system time to another time zone.
+- For the component to work as intended, the underlying platform
+  must provide the component with a means to access
+  the current system time and the system time zone.
+
 ## APIs for building WebAssembly components
+
+In general, a platform that runs components
+must provide well-defined APIs for accessing functionality
+that components need:
+for example, reading from standard input,
+accessing environment variables,
+or manipulating network sockets.
 
 It's useful to have a standard, shared set of APIs
 that WebAssembly components can depend on.
--- a/component-model/src/introduction.md
+++ b/component-model/src/introduction.md
@@ -3,6 +3,7 @@
 The WebAssembly Component Model is a broad-reaching architecture for building interoperable WebAssembly libraries, applications, and environments.
 
 This documentation is aimed at _users_ of the component model: developers of libraries and applications.
+
 > [!NOTE]
 >
 > _Compiler and Wasm runtime developers_ can take a look at the [Component Model specification](https://github.com/WebAssembly/component-model) to
--- a/component-model/src/introduction.md
+++ b/component-model/src/introduction.md
@@ -2,6 +2,14 @@
 
 The WebAssembly Component Model is a broad-reaching architecture for building interoperable WebAssembly libraries, applications, and environments.
 
+This documentation is aimed at _users_ of the component model: developers of libraries and applications.
+> [!NOTE]
+>
+> _Compiler and Wasm runtime developers_ can take a look at the [Component Model specification](https://github.com/WebAssembly/component-model) to
+> see how to add support for the component model to their project.
+
+## Table of contents
+
 | Understanding components | Building components  | Using components  |
 |--------------------------|----------------------|-------------------|
 | [Why Components?]        | [C/C++]              | [Composing]       |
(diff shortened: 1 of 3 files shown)


component-model/src/introduction.md
```diff
@@ -27,17 +35,24 @@ The WebAssembly Component Model is a broad-reaching architecture for building in
...
-> see how to add support for the component model to their project.
 
-## Status
+## APIs for building WebAssembly components
+
+It's useful to have a standard, shared set of APIs
+that WebAssembly components can depend on.
+[WASI](https://wasi.dev/) (the WebAssembly System Interface) is a standards-track specification
+that defines these APIs.
+A system or platform may expose some or all of the WASI APIs to components.
```
vados-cosmonic: I wonder if we can somewhat pivot this to focus more on "platforms", and a peek at the practical use of (ex. `wasi:cli/stdin`, `wasi:cli/environment`). 

I get the feeling that "APIs" is overloaded here and it is likely hard for newcomers to tell that these APIs are for components to use to access the underlying platform.

The [WASI README](https://github.com/WebAssembly/wasi) is pretty good here, though the first definition offered up is overly vague, the second one is decent:

> Define a set of portable, modular, runtime-independent, and WebAssembly-native APIs which can be used by WebAssembly code to interact with the outside world. These APIs preserve the essential sandboxed nature of WebAssembly through a [Capability-based](https://en.wikipedia.org/wiki/Capability-based_security) API design.

component-model/src/introduction.md
```diff
@@ -27,17 +35,24 @@ The WebAssembly Component Model is a broad-reaching architecture for building in
...
+
+### Status
 
-[WASI 0.2.0 was released](https://github.com/WebAssembly/WASI/pull/577) Jan 25, 2024, providing a stable release of WASI and the component model.
-This [is a stable set of WIT definitions](https://github.com/WebAssembly/WASI/tree/main/wasip2) that components can target. WASI proposals will
-continue to evolve and new ones will be introduced; however, users of the component model can now pin to any stable release >= `v0.2.0`. See [WASI.dev](https://wasi.dev) to stay up to date on the latest releases.
+The current stable release of WASI is [WASI 0.2.0](https://github.com/WebAssembly/WASI/pull/577),
```
vados-cosmonic, line 81: I'm not sure the original intent but do you think the older code was intended to be a sort of timeline? I'm certainly happy with this reformatting though (BIG fan of semantic line breaks!). 

Curious if you think there's any value in maintaining a sort of timeline style here (though it is likely to get out of date)

component-model/src/introduction.md
```diff
@@ -2,6 +2,14 @@
 
 The WebAssembly Component Model is a broad-reaching architecture for building interoperable WebAssembly libraries, applications, and environments.
 
+This documentation is aimed at _users_ of the component model: developers of libraries and applications.
+> [!NOTE]
+>
```
vados-cosmonic:
```suggestion
This documentation is aimed at _users_ of the component model: developers of libraries and applications.

> [!NOTE]
>
```

vados-cosmonic (approved):
LGTM 🚀 

Thanks for working on this @catamorphism it looks great so far. I left a couple nits/questions but these are more worth pondering on.

component-model/src/introduction.md
catamorphism: Added some more detail in [2da3ade](https://github.com/bytecodealliance/component-docs/pull/280/commits/2da3ade84e6f3b8d2f53d05e5e6be3a182357d82), I agree that the use of "APIs" is overloaded; let me know if you think this clarifies!

component-model/src/introduction.md
catamorphism: Hmm, I'm not sure if it was supposed to be a timeline, since it doesn't mention when WASI 0.1 was released (and I don't think it needs to in an introduction). I think what this needs to convey is what the stable version is and where to follow future versions, but if there's anything else you think should be there, let me know.

vados-cosmonic: Yeah, I'm happy with the stuff that's there now, thanks!

component-model/src/introduction.md
vados-cosmonic: This is definitely a big improvement, thanks!

component-model/src/introduction.md
itowlson: I think the original intent here was: at the time, WASI was in transition.  Existing content around the web largely referred to WASI 0.1 (without the versioning), and/or various drafts and snapshots.  So there was a desire to provide some context to "this is what we are talking about here, this is the timestamp by which to gauge how tools or other posts might or might not reflect it, this is where to go to see what it looks like, this is how it is intended to interact with the future."

The context for readers today is hopefully different, as WASI 0.2 content and tooling has rolled out, and WASI 0.1 and 0.2 are now more clearly distinguished.  So it's certainly reasonable to reappraise what we need to say here.  I don't have an opinion on whether to break it up into multiple paragraphs (for me it reads a bit staccato, disconnecting the thoughts, so I guess I do have an opinion after all, but it's not a strong one).

I would observe that if we are documenting the "current stable release" then that is probably WASI 0.2.2 or 0.2.3 or something, I haven't been following in detail _grin_

Also, excellent username, Tim!

itowlson: LITERALLY SECONDS TOO LATE

catamorphism: @itowlson 

> I would observe that if we are documenting the "current stable release" then that is probably WASI 0.2.2 or 0.2.3 or something, I haven't been following in detail _grin_
> 

Ah, good point. Since I plan to review the entire book including the code examples, maybe the best thing to do would be to verify that the examples work with 0.2.5 and then revise this text to say that the contents of the book reflect 0.2.5, without saying that this is the latest stable version?

> Also, excellent username, Tim!

Thanks :)

itowlson: > Since I plan to review the entire book including the code examples, maybe the best thing to do would be to verify that the examples work with 0.2.5 and then revise this text to say that the contents of the book reflect 0.2.5

I'd guess that nothing in the book is dependent on WASI 0.2.5 features, as (from rather dusty memory) the book focuses on the underlying component model and uses WASI only incidentally (e.g. when exercising a sample in the CLI).  At the same time:

* it's not worth digging out some ancient Wasmtime that only has 0.2.0 and running tests with that
* there may be some post-0.2.0 features at the WIT/CM level, although the only one I can think of is `@since/@unstable` in WIT
* I am not sure how we even refer to post-0.2.0 modifications to WIT and the CM

So... uh... I guess... verifying the examples using a WASI 0.2.5 implementation sounds fine, but I am not sure what to do about any "contents reflect" remark - it may need to be post-0.2 if you decide to update the WIT docs, but I'd avoid implying "if your toolchain is more than three months old, you're on your own."  Sorry not to offer anything more concrete!

vados-cosmonic: Oh and by the way ya'll... the latest might even be 0.2.6... https://github.com/WebAssembly/wasi-cli/commit/939bd6d492c11bbd7d3c349b91096061022bc3d7

Though of course, the upstream docs aren't quite updated yet :)

PR #281, opened 2025-07-16
Revise and reorganize "Why the Component Model?" section

catamorphism:
I've tried to revise this section with a Wasm beginner in mind, as well as breaking it up into sections so it's a little easier to skim.

In some cases it was hard to be sure how much detail to go into vs. linking to external sources (e.g. memory), but I tried to strike a balance between not intimidating the reader with too much detail, and not intimidating the reader with undefined terms.

The diffs are a bit hard to follow, so it's probably easiest to review by viewing the rendered Markdown side-by-side with the current version.

### DIFF
--- a/component-model/src/design/why-component-model.md
+++ b/component-model/src/design/why-component-model.md
@@ -31,7 +31,8 @@ A core module is a set of definitions.
 Kinds of definitions include:
 * _Functions_ define executable units of code
   (sequences of instructions along with declarations
-  for argument names and types and return types).
+  for the names of arguments
+  and the types of arguments and return values).
 * [_Linear memories_](https://webassembly.github.io/spec/core/syntax/modules.html#syntax-mem)
   define buffers of uninterpreted bytes that can be read from
   and written to by instructions.
@@ -199,7 +200,7 @@ and thus it cannot indirectly communicate to others
 by writing to its memory and having others read from that memory.
 This not only reinforces sandboxing, but enables interoperation
 between languages that make different assumptions about memory:
-for example, allowing a component that relies garbage-collected memory
+for example, allowing a component that relies on garbage-collected memory
 to interoperate with one that uses conventional linear memory.
 
 ## Using components


component-model/src/design/why-component-model.md
```diff
@@ -1,19 +1,126 @@
 # Why the Component Model?
 
-If you've tried out WebAssembly, you'll be familiar with the concept of a _module_. Roughly speaking, a module corresponds to a single `.wasm` file, with functions, memory, imports and exports, and so on. These "core" modules can run in the browser, or via a separate runtime such as Wasmtime or WAMR. A module is defined by the [WebAssembly Core Specification](https://webassembly.github.io/spec/core/), and if you compile a program written in Rust, C, Go or whatever to WebAssembly, then a core module is what you'll get.
+WebAssembly programs can be written by hand,
```
vados-cosmonic, line 16: So I actually like the original starting text a bit more here -- the original text was *very* clear about what a module was, then the actual benefit of the component model. 

This (new) section explains really clearly how to create a WebAssembly core module/how modules are created, but kind of leaves out the namesake of "why" the component model is a good idea/where it fits in. 

Maybe not a full on revert, but is there a way we can front-load the "core modules can only interact with the outside world in a way that becomes tedious without high level types" and move all the core module specific stuff to the newly created core module section? 

Maybe this sort of... pre-section can start with text like "At a high level, the component model builds upon WebAssembly core modules in order to ..."

component-model/src/design/why-component-model.md
```diff
@@ -1,19 +1,126 @@
...
 
-Core modules are, however, limited in how they expose their functionality to the outside world to functions that take and return only a small number of core WebAssembly types (essentially only integers and floating-point numbers). Richer types, such as strings, lists, records (a.k.a. structs), etc. have to be represented in terms of integers and floating point numbers, for example by the use of pointers and offsets. Those representations are often times not interchangeable across languages. For example, a string in C might be represented entirely differently from a string in Rust or in JavaScript.
+## Core modules
 
-For Wasm modules to interoperate, therefore, there needs to be an agreed-upon way for exposing those richer types across module boundaries.
+Typically, a core module defines a set of _functions_
+along with auxiliary definitions
+that are necessary for executing those functions.
+Functions are made up of _instructions_.
+Auxiliary definitions include:
```
vados-cosmonic: IMO functions should be just another item in the list here -- I'm not sure they deserve to be *so* prominently features. 

Also, is it worth linking to the WebAssembly core spec here somewhere?

component-model/src/design/why-component-model.md
```diff
@@ -1,19 +1,126 @@
...
+but it's more likely that you will use a compiler to generate your programs.
+In general, a compiler translates programs from a source language
+to a target language.
+Compilers whose target language is WebAssembly may take
+Rust, C, Go, or a variety of other languages as a source language.
+In this case, the compiler produces a WebAssembly _core module_.
+Core modules have some limitations, which components address.
 
-Core modules are, however, limited in how they expose their functionality to the outside world to functions that take and return only a small number of core WebAssembly types (essentially only integers and floating-point numbers). Richer types, such as strings, lists, records (a.k.a. structs), etc. have to be represented in terms of integers and floating point numbers, for example by the use of pointers and offsets. Those representations are often times not interchangeable across languages. For example, a string in C might be represented entirely differently from a string in Rust or in JavaScript.
+## Core modules
```
vados-cosmonic:
```suggestion
## WebAssembly Core modules
```

Maybe this is implicit/redundant?

component-model/src/design/why-component-model.md
```diff
@@ -1,19 +1,126 @@
...
 
-For Wasm modules to interoperate, therefore, there needs to be an agreed-upon way for exposing those richer types across module boundaries.
+Typically, a core module defines a set of _functions_
+along with auxiliary definitions
+that are necessary for executing those functions.
+Functions are made up of _instructions_.
+Auxiliary definitions include:
+* _Linear memories_ define untyped buffers that can be read from
```
vados-cosmonic: A link to the [memories section](https://webassembly.github.io/spec/core/syntax/modules.html#syntax-mem) may be reasonable here -- also I think these buffers *have* to be bytes (or put another way I'm not sure they're "untyped"/if I am parsing that correctly).

component-model/src/design/why-component-model.md
```diff
@@ -1,19 +1,126 @@
...
+   the functions in the module,
+   along with type signatures for required functions
+   in the imported module.
+* _Exports_ define the names of functions within
+  the module that should be accessible externally.
+* And others; see [the Core Specification](https://webassembly.github.io/spec/core/syntax/modules.html)
+  for the complete list.
 
-In the component model, these type definitions are written in a language called [WIT (Wasm Interface Type)](./wit.md), and the way they translate into bits and bytes is called the [Canonical ABI (Application Binary Interface)](./../advanced/canonical-abi.md). A Wasm [component](./components.md) is thus a wrapper around a core module that specifies its imports and exports using such [Interfaces](./interfaces.md).
+A core module usually corresponds to a single `.wasm` file.
```
vados-cosmonic:
```suggestion
A core module usually corresponds to a single `.wasm` binary.
```

We have `file` support, *maybe* that example output is worth putting here, so people can identify a `.wasm` file?

That might also pay off in a section describing components, as the `file` output there is different

component-model/src/design/why-component-model.md
```diff
@@ -1,19 +1,126 @@
...
+These modules can be run in the browser,
+or via a separate runtime such as [Wasmtime](https://wasmtime.dev/)
```
vados-cosmonic, line 50:
```suggestion
or via a server-side WebAssembly runtime such as [Wasmtime](https://wasmtime.dev/)
```

component-model/src/design/why-component-model.md
```diff
@@ -1,19 +1,126 @@
...
+or [WAMR](https://github.com/bytecodealliance/wasm-micro-runtime).
 
-The agreement of an interface adds a new dimension to Wasm portability. Not only are components portable across architectures and operating systems, but they are now portable across languages. A Go component can communicate directly and safely with a C or Rust component. It need not even know which language another component was written in - it needs only the component interface, expressed in WIT. Additionally, components can be linked into larger graphs, with one component's exports satisfying another's imports.
+A module is defined by the [WebAssembly Core Specification](https://webassembly.github.io/spec/core/).
```
vados-cosmonic: Feels like this should be higher up, maybe even the opening line of this section?

component-model/src/design/why-component-model.md
```diff
@@ -1,19 +1,126 @@
...
+### Limitations of core modules
 
-Moreover, a component interacts with a runtime or other components _only_ by calling its imports and having its exports called. Specifically, unlike core modules, a component may not export Wasm memory, and thus it cannot indirectly communicate to others by writing to its memory and having others read from that memory. This not only reinforces sandboxing, but enables interoperation between languages that make different assumptions about memory - for example, allowing a component that relies on Wasm GC (garbage collected) memory to collaborate with one that uses conventional linear memory.
+Core modules are, however, limited in how they expose their functionality to the outside world.
+Core modules expose their functionality by exporting functions.
```
vados-cosmonic:
```suggestion
Core modules expose their functionality by exporting functions.
```
I don't think this changes with respect to components, so maybe not worth mentioning? Or was the idea to contrast this with interfaces exported w/ the component model?

component-model/src/design/why-component-model.md
```diff
@@ -1,19 +1,126 @@
...
+These functions' argument and return types are restricted, essentially,
```
vados-cosmonic:
```suggestion
WebAssembly core function argument and return types are restricted, essentially,
```

component-model/src/design/why-component-model.md
```diff
@@ -1,19 +1,126 @@
...
+to only integers and floating-point numbers.
```
vados-cosmonic:
```suggestion
to only integers and floating-point numbers (`i32`, `i64`, `f32`, `f64`).
```
NIT: I personally like introducing the types concretely like this when explaining the basics of core modules

component-model/src/design/why-component-model.md
```diff
@@ -1,19 +1,126 @@
...
+Compound types, such as strings, lists, arrays, records, or structs,
+have to be represented in terms of integers and floating-point numbers.
```
vados-cosmonic:
```suggestion
Compound types common in higher level programming langauges such as strings, lists, arrays, enums, or structs,
have to be represented in terms of integers and floating-point numbers.
```

component-model/src/design/why-component-model.md
```diff
@@ -1,19 +1,126 @@
...
+
+Recall that a linear memory is an uninitialized region of bytes
+declared within a module.
+So, a string argument might be represented as two separate arguments:
+an integer offset into a memory,
```
vados-cosmonic: An example function signature might illuminate this more (and it might be worth spacing this out a bit) -- whether WAT or in some pseudo code?

component-model/src/design/why-component-model.md
```diff
@@ -1,19 +1,126 @@
...
+and an integer representing the length of the string.
+These representations are frequently specific to each programming language.
+For example, a string in C might be represented entirely differently
+from a string in Rust or in JavaScript.
```
vados-cosmonic: I think we can say this with certainty -- strings in all three languages actually *are* represented differently/have different constraints.

component-model/src/design/why-component-model.md
```diff
@@ -1,19 +1,126 @@
...
+Moreover, to make this approach work, modules must import and export memories,
+which can be error-prone, as different languages
+make different assumptions about memory layout.
+
+For WebAssembly modules to interoperate, therefore, there needs to be an agreed-upon way
```
vados-cosmonic:
```suggestion
For WebAssembly modules written in different languages to interoperate smoothly, there needs to be an agreed-upon way
```

component-model/src/design/why-component-model.md
```diff
@@ -1,19 +1,126 @@
...
+to expose these richer types across module boundaries.
+
+## Components
+
+Components were developed to ease interoperability between modules.
```
vados-cosmonic: Components solve at least two problems we've seen so far; rich types and cross-language interop. What do you think about making a reference to those previous motivating sections and maybe making this a list? 

In my head I'm thinking something like "Components help address both of the aforementioned problems with using ad-hoc WebAssembly core components:" ?

component-model/src/design/why-component-model.md
```diff
@@ -1,19 +1,126 @@
...
+Conceptually, a component is a module that is restricted
+to interact only through its imported and exported functions.
+Compared to core modules, components also use a richer
+mechanism for expressing the types of functions.
+
+### Interfaces
+
+These interfaces are expressed in a separate language called [WIT (Wasm Interface Type)](./wit.md).
```
vados-cosmonic:
```suggestion
These interfaces are expressed in a separate language called [WebAssembly Interface types (WIT)](./wit.md).
```

component-model/src/design/why-component-model.md
```diff
@@ -1,19 +1,126 @@
...
+Interfaces contain _types_.
```
vados-cosmonic: types, and functions, maybe worth linking directly to [the WIT section](https://github.com/WebAssembly/component-model/blob/main/design/mvp/WIT.md#wit-interfaces) here, or to our WIT explainer

component-model/src/design/why-component-model.md
```diff
@@ -1,19 +1,126 @@
...
+The bit-level representations of these types are specified by
+the [Canonical ABI (Application Binary Interface)](./../advanced/canonical-abi.md).
+
+### Interoperability
+
+Interfaces make it possible to write components that are
+portable across different architectures and operating systems.
```
vados-cosmonic: I'm not sure this is true -- it's more that they enable bundling different types and functionality in high level languages which can be used across components. 

I think WebAssembly core accomplishes these goals inherently

component-model/src/design/why-component-model.md
```diff
@@ -1,19 +1,126 @@
...
+Not only that, components are portable across different programming languages.
+A component implemented in Go can communicate directly and safely
+with a C or Rust component.
```
vados-cosmonic:
```suggestion
with a C or Rust component, by relying on the shared conventions of the Component Model ABI.
```

component-model/src/design/why-component-model.md
```diff
@@ -1,19 +1,126 @@
...
+Writing a component doesn't even require knowledge
+of which language its dependent components are implemented in,
+only the component interface expressed in WIT.
+Additionally, components can be linked into larger graphs,
+with one component's exports satisfying another's imports.
```
vados-cosmonic:
```suggestion
Additionally, components can be composed into larger graphs,
with one component's exports satisfying another's imports.
```

Good time to link to https://component-model.bytecodealliance.org/composing-and-distributing/composing.html IMO

component-model/src/design/why-component-model.md
```diff
@@ -1,19 +1,126 @@
...
+
+### Benefits of the component model
+
+Putting all of the pieces together:
+the component model is a way of writing WebAssembly modules
+that interact with each other only through exports and imports of functions
+whose types are expressed using WIT.
+
+When combined with Wasm's strong [sandboxing](https://webassembly.org/docs/security/),
```
vados-cosmonic: I think this wording might be a little weird, because the component model can't exist *without* Wasm as the lower layer essentially... So more like "building upon" than "combined with"?

component-model/src/design/why-component-model.md
```diff
@@ -1,19 +1,126 @@
...
+the component model has further benefits.
+Richer type signatures express richer semantic properties
+than type signatures made up only of integers and floats.
+Rich types make it possible to statically analyse
+and reason about a component's behaviour.
+Simply by examining the surface of the component—the types
+of its imports and exports—properties can be
+enforced and guaranteed.
```
vados-cosmonic: Do you think it might be a good idea to put an example WIT here to drive the point home? Might help to also break up the prose here

component-model/src/design/why-component-model.md
```diff
@@ -1,19 +1,126 @@
...
+between languages that make different assumptions about memory:
+for example, allowing a component that relies garbage-collected memory
+to interoperate with one that uses conventional linear memory.
+
+## Using components
 
 Now that you have a better idea about how the component model can help you, take a look at [how to build components](../language-support.md) in your favorite language!
 
+## Further reading
+
```
vados-cosmonic, line 211: Now that this is a whole section, we can probably remove the `>` ?

I'm actually not even sure why it's a quote like that -- probably was intended to highlight?

vados-cosmonic (commented):
Hey thanks for working on this @catamorphism -- I had a bunch of comments but hopefully they all make sense/happy to go through them synchronously at some point as well!

component-model/src/design/why-component-model.md
catamorphism: I see your point -- see what you think about [9f86658](https://github.com/bytecodealliance/component-docs/pull/281/commits/9f86658d3e0e237cdae8e5c499db928913fd45b9)?

The original text does start with quite a bit of explanation about core modules as well. By defining core modules first and then moving to components, I was hoping to avoid having to add a connecting sentence like "To understand what the limitations of core modules are, we start by defining them." But I ended up doing exactly that. I don't like this kind of sentence, but maybe you can think of a way to avoid it.

component-model/src/design/why-component-model.md
catamorphism: Changed in [0c2edc5](https://github.com/bytecodealliance/component-docs/pull/281/commits/0c2edc503abc6131618aa014e8b717ca229dfd71)

component-model/src/design/why-component-model.md
catamorphism: Changed in [0c2edc5](https://github.com/bytecodealliance/component-docs/pull/281/commits/0c2edc503abc6131618aa014e8b717ca229dfd71)

Good point, "untyped" wasn't quite right. Changed to "buffers of uninterpreted bytes".

component-model/src/design/why-component-model.md
catamorphism: Changed to "binary `.wasm` file" to cover all bases.

I'm not sure what you mean by example output -- do you mean showing the binary contents of a `.wasm` file or the text-based representation of it that a tool would print out?

component-model/src/design/why-component-model.md
catamorphism: Changed in [0c2edc5](https://github.com/bytecodealliance/component-docs/pull/281/commits/0c2edc503abc6131618aa014e8b717ca229dfd71)

component-model/src/design/why-component-model.md
catamorphism: Good point, I rephrased in [17b64a4](https://github.com/bytecodealliance/component-docs/pull/281/commits/17b64a492b6292100a9ebbfb0b7df0d203f4eb4a).

component-model/src/design/why-component-model.md
catamorphism: Changed in [17b64a4](https://github.com/bytecodealliance/component-docs/pull/281/commits/17b64a492b6292100a9ebbfb0b7df0d203f4eb4a) (slightly different phrasing)

component-model/src/design/why-component-model.md
catamorphism: Changed in [17b64a4](https://github.com/bytecodealliance/component-docs/pull/281/commits/17b64a492b6292100a9ebbfb0b7df0d203f4eb4a) (slightly different phrasing)

component-model/src/design/why-component-model.md
catamorphism: Changed in [17b64a4](https://github.com/bytecodealliance/component-docs/pull/281/commits/17b64a492b6292100a9ebbfb0b7df0d203f4eb4a) (slightly different phrasing)

component-model/src/design/why-component-model.md
catamorphism: Good idea, added in [17b64a4](https://github.com/bytecodealliance/component-docs/pull/281/commits/17b64a492b6292100a9ebbfb0b7df0d203f4eb4a)

component-model/src/design/why-component-model.md
catamorphism: Changed in [17b64a4](https://github.com/bytecodealliance/component-docs/pull/281/commits/17b64a492b6292100a9ebbfb0b7df0d203f4eb4a)

component-model/src/design/why-component-model.md
catamorphism: See what you think of [5e65915](https://github.com/bytecodealliance/component-docs/pull/281/commits/5e65915ed87c1d8acd3aef90895fb9f6dcb81a09)?

component-model/src/design/why-component-model.md
catamorphism: Hmm, I've preferred using "WebAssembly" over "Wasm" for consistency, but in https://github.com/WebAssembly/component-model/blob/main/design/mvp/WIT.md , WIT is introduced as "The Wasm Interface Type (WIT) format". Should it be different here?

component-model/src/design/why-component-model.md
catamorphism: Changed in [005f1a6](https://github.com/bytecodealliance/component-docs/pull/281/commits/005f1a652211cdbd397759678732a0c4d15ffa24)

component-model/src/design/why-component-model.md
catamorphism: Good point, revised this in [0aea4b3](https://github.com/bytecodealliance/component-docs/pull/281/commits/0aea4b327e477c26a059653918d3efde4eeae828)

component-model/src/design/why-component-model.md
catamorphism: I think including WIT might be jumping ahead -- there would need to be a lot of explanation of the syntax that doesn't seem to fit into a persuasive mode. A diagram might help; I'm not the greatest with diagrams, but maybe I can think about that after the text is in agreed-upon shape.

component-model/src/design/why-component-model.md
catamorphism: Fixed in [262cec0](https://github.com/bytecodealliance/component-docs/pull/281/commits/262cec0552e319c517020d165626db81dc9dcd08)

component-model/src/design/why-component-model.md
catamorphism: Changed in [262cec0](https://github.com/bytecodealliance/component-docs/pull/281/commits/262cec0552e319c517020d165626db81dc9dcd08)

catamorphism:
I think I've addressed all the review comments now. @vados-cosmonic let me know if I missed anything!

component-model/src/design/why-component-model.md
vados-cosmonic: Thanks, that's perfect!

component-model/src/design/why-component-model.md
vados-cosmonic: Ahhh sorry I was unclear there -- I meant outputting what `file some.wasm` would put out. I'm definitely flexible on that but I think it would help people build some debugging intuition...

component-model/src/design/why-component-model.md
vados-cosmonic: Thanks!

component-model/src/design/why-component-model.md
```diff
@@ -39,32 +39,55 @@ Kinds of definitions include:
...
+```
+remove-duplicates: func(offset: i32, length: i32) -> [i32, i32]
+```
+
+supposing that `remove-duplicates` is a function
+to create a new string consisting of the unique characters
+in its argument.
+The return type is a pair of 32-bit integers,
+reflecting that the function must return
+the new offset for the newly allocated string, as well as its length.
```
vados-cosmonic: This is a *great* example. I'm not sure what I expected, maybe string reverse or something, but this is pretty useful seeming/realistic.

The only thing I'd add here is that we should explain the args as skillfully as we explain the return type. I know it's down in the weeds but I think actually showing how painful it is will motivate even more effectively.

Right now for the return argument we have this:

> The return type is a pair of 32-bit integers,
> reflecting that the function must return
> the new offset for the newly allocated string, as well as its length.

Maybe this could be something like:

> The return type is a list of integers, the first being an offset into the linear memory of the
> component -- where the newly allocated string starts -- along with a length.
> After calling this function, the caller has to reach into the linear memory and attempt to read
> the created string out, given the provided offset and length.

component-model/src/design/why-component-model.md
```diff
@@ -39,32 +39,55 @@ Kinds of definitions include:
...
 an integer offset into a memory,
 and an integer representing the length of the string.
-These representations are frequently specific to each programming language.
-For example, a string in C might be represented entirely differently
+
+In pseudocode, a type signature for a string-manipulating function
+might look like:
+
+```
+remove-duplicates: func(offset: i32, length: i32) -> [i32, i32]
```
vados-cosmonic, line 76: If we're going for valid WIT `[i32,i32]` -> `list<i32>` or `list<i32, 2>` or `tuple<i32, i32>` 

In reality, this is even more painful because it'd be something like `i64` and then relevant packing/unpacking (or however someone decided to do it!)

[EDIT] note that if we're going for just pure pseudo code we can use a recognizable syntax rather than WIT syntax, JS or something like that? 

I'm flexible here, but almost-WIT was a bit jarring (thought it would be awesome if we had that syntax!

[EDIT] You know what, this is fine -- I think it's easy to parse/grok

component-model/src/design/why-component-model.md
vados-cosmonic: I think it's good! I hesitated to start the section with a back-reference (in case someone skipped around), but I think it definitely lays out those two benefits well. If someone stopped reading after that sentence, they'd at least pull the right intuition away.

component-model/src/design/why-component-model.md
vados-cosmonic: Thanks this is a good point -- Wasm is a shortening of WebAssembly so I'm not sure it matters a ton -- for example [`wit-bindgen` uses "WebAssembly Interface Types"](https://github.com/bytecodealliance/wit-bindgen), I *think* the words should be interchangable.

That said, consistency is better, so I'm happy with Wasm Interface Types! What I really wanted to point out was to have the acronym in parens and the full phrase first, at least that's my personal preference

component-model/src/design/why-component-model.md
vados-cosmonic: Yup, fair point -- it's a lot to try and jam in! I don't think you have to try and add a diagram here YET.

component-model/src/design/why-component-model.md
vados-cosmonic: Hey @catamorphism I think this one might have gotten lost in the shuffle, what do you think? Also open to "Wasm" here over WebAssembly

component-model/src/design/why-component-model.md
```diff
@@ -1,19 +1,185 @@
...
-For Wasm modules to interoperate, therefore, there needs to be an agreed-upon way for exposing those richer types across module boundaries.
+A module is defined by the [WebAssembly Core Specification](https://webassembly.github.io/spec/core/).
 
-In the component model, these type definitions are written in a language called [WIT (Wasm Interface Type)](./wit.md), and the way they translate into bits and bytes is called the [Canonical ABI (Application Binary Interface)](./../advanced/canonical-abi.md). A Wasm [component](./components.md) is thus a wrapper around a core module that specifies its imports and exports using such [Interfaces](./interfaces.md).
+WebAssembly programs can be written by hand,
+but it's more likely that you will use a compiler to generate your programs.
+In general, a compiler translates programs from a source language
+to a target language.
+Compilers whose target language is WebAssembly may take
+Rust, C, Go, or a variety of other languages as a source language.
```
vados-cosmonic:
```suggestion
but it's more likely that you will use a higher level programming language
such as Rust, C, Go, Javascript, or Python to build WebAssembly programs.
```

What do you think about this? I figured we didn't need to try and explain compilers, and that actually paradoxically feels like leaving out interpreted languages we support.

component-model/src/design/why-component-model.md
```diff
@@ -1,19 +1,185 @@
...
+In this case, the compiler produces a WebAssembly [core module](https://webassembly.github.io/spec/core/syntax/modules.html).
```
vados-cosmonic:
```suggestion
Many existing toolchains currently produce a [WebAssembly core module](https://webassembly.github.io/spec/core/syntax/modules.html) -- a single binary `.wasm` file.
```

component-model/src/design/why-component-model.md
```diff
@@ -1,19 +1,185 @@
...
+   the functions in the module,
+   along with type signatures for required functions
+   in the imported module.
+* _Exports_ define the names of functions within
+  the module that should be accessible externally.
+* And others; see [the Core Specification](https://webassembly.github.io/spec/core/syntax/modules.html)
+  for the complete list.
 
-Combined with Wasm's strong sandboxing, this opens the door to yet further benefits. By expressing higher-level semantics than integers and floats, it becomes possible to statically analyse and reason about a component's behaviour - to enforce and guarantee properties just by looking at the surface of the component. The relationships within a graph of components can be analysed, for example to verify that a component containing business logic has no access to a component containing personally identifiable information.
+A core module usually corresponds to a single binary `.wasm` file.
```
vados-cosmonic:
```suggestion
```

This line may not be needed if you move it up above (earlier comment)

component-model/src/design/why-component-model.md
```diff
@@ -1,19 +1,185 @@
...
+These modules can be run in the browser,
+or via a separate runtime such as [Wasmtime](https://wasmtime.dev/)
+or [WAMR](https://github.com/bytecodealliance/wasm-micro-runtime).
 
-Moreover, a component interacts with a runtime or other components _only_ by calling its imports and having its exports called. Specifically, unlike core modules, a component may not export Wasm memory, and thus it cannot indirectly communicate to others by writing to its memory and having others read from that memory. This not only reinforces sandboxing, but enables interoperation between languages that make different assumptions about memory - for example, allowing a component that relies on Wasm GC (garbage collected) memory to collaborate with one that uses conventional linear memory.
+### Limitations of core modules
+
+Core modules are, however, limited in how they expose their functionality to the outside world.
```
vados-cosmonic:
```suggestion
Core modules are limited in the computation they can perform and 
how they expose their functionality to the outside world.
```

component-model/src/design/why-component-model.md
```diff
@@ -1,19 +1,185 @@
...
+and only these types can be returned from functions as results.
+Compound types common in higher-level programming languages,
+such as strings, lists, arrays, enums (enumerations), or structs (records),
+have to be represented in terms of integers and floating-point numbers.
+
+Recall that a linear memory is an uninitialized region of bytes
+declared within a module.
+So, a string argument might be represented as two separate arguments:
+an integer offset into a memory,
+and an integer representing the length of the string.
```
vados-cosmonic: What do you think about swapping these sentences and starting with "For example"

So something like:

> For example, for a function to accept a string, ... into a linear memory.
> Recall that a linear memory is an ...

component-model/src/design/why-component-model.md
vados-cosmonic: You know what, this is fine -- I think it's easy to parse/grok

component-model/src/design/why-component-model.md
```diff
@@ -1,19 +1,185 @@
...
+
+We would prefer to write a pseudocode type signature like this:
+
+```
+remove-duplicates: func(s: string) -> string
+```
+
+and dispense with the memory exports and imports altogether.
+
+Data representations are frequently specific to each programming language.
```
vados-cosmonic, line 128: We could maybe enhance this just slightly with a phrase like "The complexity doesn't stop there", because just getting the bytes over is one thing, properly interpreting them in the receiver is another!

component-model/src/design/why-component-model.md
```diff
@@ -1,19 +1,185 @@
...
+
+For WebAssembly modules written in different languages to interoperate smoothly,
+there needs to be an agreed-upon way to expose these richer types across module boundaries.
+
+## Components
+
+Components solve the two problems that we've seen so far:
+the limited type system of core module functions,
+and cross-language interoperability.
+Conceptually, a component is a module that is restricted
```
vados-cosmonic:
```suggestion
Conceptually, a component is a WebAssembly binary (which may or may not contain modules) that is restricted
```

component-model/src/design/why-component-model.md
```diff
@@ -1,19 +1,185 @@
...
+to interact only through its imported and exported functions.
+Compared to core modules, components also use a richer
+mechanism for expressing the types of functions.
```
vados-cosmonic:
```suggestion
Compared to core modules, components use a different binary format and opt-in to a richer
mechanism for expressing the types of functions by default.
```

component-model/src/design/why-component-model.md
```diff
@@ -1,19 +1,185 @@
...
+[Interfaces](./wit.md#interfaces) contain definitions of _types_
+and type signatures for [_functions_](./wit.md#functions).
+The bit-level representations of types are specified by
+the [Canonical ABI (Application Binary Interface)](./../advanced/canonical-abi.md).
+
+### Interoperability
+
+WebAssembly core modules are already portable across different architectures
+and operating systems;
+components retain these benefits and, using interfaces,
```
vados-cosmonic:
```suggestion
components retain these benefits and, using the Component Model ABI,
```

component-model/src/design/why-component-model.md
```diff
@@ -1,19 +1,185 @@
...
+add portability across different programming languages.
+A component implemented in Go can communicate directly and safely
+with a C or Rust component, by relying on the shared conventions of the Component Model ABI.
+Writing a component doesn't even require knowledge
+of which language its dependent components are implemented in,
+only the component interface expressed in WIT.
+Additionally, components can be [composed](../composing-and-distributing.md) into larger graphs,
+with one component's exports satisfying another's imports.
+
+### Benefits of the component model
```
vados-cosmonic, line 176: Do we want "Component Model" in most places?

component-model/src/design/why-component-model.md
```diff
@@ -1,19 +1,185 @@
...
+
+Putting all of the pieces together:
+the component model is a way of writing WebAssembly modules
+that interact with each other only through exports and imports of functions
+whose types are expressed using WIT.
```
vados-cosmonic:
```suggestion
the component model is a binary WebAssembly format encapsulates WebAssembly
modules and enables interactions across modules through exports and imports of functions
whose types are expressed using WIT.
```

vados-cosmonic (commented):
Hey @catamorphism thanks for the fixups -- I took another look through.

I think it's really important to get this right, thanks for the hard work here

component-model/src/design/why-component-model.md
```diff
@@ -1,19 +1,185 @@
 # Why the Component Model?
 
-If you've tried out WebAssembly, you'll be familiar with the concept of a _module_. Roughly speaking, a module corresponds to a single `.wasm` file, with functions, memory, imports and exports, and so on. These "core" modules can run in the browser, or via a separate runtime such as Wasmtime or WAMR. A module is defined by the [WebAssembly Core Specification](https://webassembly.github.io/spec/core/), and if you compile a program written in Rust, C, Go or whatever to WebAssembly, then a core module is what you'll get.
+At a high level, the component model builds upon WebAssembly _core modules_
+to enhance interoperability, both by enriching the type system
```
lukewagner:
```suggestion
to enhance interoperability between languages and libraries, both by enriching the type system
```

component-model/src/design/why-component-model.md
```diff
@@ -1,19 +1,185 @@
...
+used for checking the safety of interactions between modules,
+and by removing potentially error-prone ways for modules to interact.
```
lukewagner:
```suggestion
and by clearly defining and enforcing the low-level calling contract between separately-compiled modules.
```

component-model/src/design/why-component-model.md
lukewagner: "server-side" might be too specific, since Wasmtime/WAMR are used in various client and embedded devices too.

component-model/src/design/why-component-model.md
```diff
@@ -1,19 +1,185 @@
...
+Core modules are, however, limited in how they expose their functionality to the outside world.
+In WebAssembly core modules, functions are restricted, essentially,
+to using integer (`i32` or `i64`) or floating-point (`f32` or `f64`) types.
+Only these types can be passed as arguments to functions,
+and only these types can be returned from functions as results.
+Compound types common in higher-level programming languages,
+such as strings, lists, arrays, enums (enumerations), or structs (records),
+have to be represented in terms of integers and floating-point numbers.
+
+Recall that a linear memory is an uninitialized region of bytes
```
lukewagner:
```suggestion
Recall that a linear memory is an uninterpreted array of bytes
```

component-model/src/design/why-component-model.md
```diff
@@ -1,19 +1,185 @@
...
+whose types are expressed using WIT.
+
+Building upon Wasm's strong [sandboxing](https://webassembly.org/docs/security/),
+the component model has further benefits.
+Richer type signatures express richer semantic properties
+than type signatures made up only of integers and floats.
+Rich types make it possible to statically analyse
+and reason about a component's behaviour.
+Simply by examining the surface of the component—the types
+of its imports and exports—properties can be
```
lukewagner:
```suggestion
Rich types make it easier to know what a component or interface is 
doing at a glance and have guarantees of what bad things cannot happen.
```

lukewagner (commented):
Great to see the work on continually improving the docs!  Just a few drive-by suggestions:

component-model/src/design/why-component-model.md
vados-cosmonic: Good point! I'll resolve this then and we can avoid making this change

component-model/src/design/why-component-model.md
catamorphism: Ah, got it. Let me know if this is what you had in mind: [873fb8e](https://github.com/bytecodealliance/component-docs/pull/281/commits/873fb8ee1d48c7fae51629b2aee4892068e75002)

component-model/src/design/why-component-model.md
catamorphism: Thanks! Yes, I was going to use string reverse, but realized it's more motivating to use an example where the output string length can be different from the input length.

catamorphism: See what you think of [11153b2](https://github.com/bytecodealliance/component-docs/pull/281/commits/11153b28e459bb800d81b13924d7f8d8d6347bb6)?

component-model/src/design/why-component-model.md
catamorphism: I'm fine with using "WebAssembly" here if both are used elsewhere, so I changed it to what you suggested ( [117804e](https://github.com/bytecodealliance/component-docs/pull/281/commits/117804e6f61c79c1def68d165d566b276f615329) ).

component-model/src/design/why-component-model.md
catamorphism: Addressed in [428763e](https://github.com/bytecodealliance/component-docs/pull/281/commits/428763eeee8d98570aea44fe9910986beccd7b41)

component-model/src/design/why-component-model.md
catamorphism: Good idea -- done in [bfe929b](https://github.com/bytecodealliance/component-docs/pull/281/commits/bfe929b6e611c4459b712d96343ebc13c35e3ecd)

component-model/src/design/why-component-model.md
catamorphism: Also a good idea -- done in [bfe929b](https://github.com/bytecodealliance/component-docs/pull/281/commits/bfe929b6e611c4459b712d96343ebc13c35e3ecd)

component-model/src/design/why-component-model.md
catamorphism: Adopted the suggestion, but it's more accurate to say that the binary interacts through the contained _modules'_ imports and exports, right?

component-model/src/design/why-component-model.md
catamorphism: Adopted suggestion (with a little editing) -- I thought about including the `file` output here to contrast with the earlier example of `file` output, but it doesn't seem too helpful, since it doesn't explicitly mention components:

```
$ file add.component.wasm 
add.component.wasm: WebAssembly (wasm) binary module version 0x1000d
```

What do you think?

component-model/src/design/why-component-model.md
catamorphism: Do you mean whether to capitalize "Component Model" or not? I have no strong feelings as long as it's consistent. (Or do you mean whether to say "component model" or "components"? Since the section is headed 'why the component model?' I'm more inclined to use 'component model' at least in this section...)

component-model/src/design/why-component-model.md
catamorphism: Did this a little differently in [078c562](https://github.com/bytecodealliance/component-docs/pull/281/commits/078c5622bc7dcce3faca191b4a345b4452a17dd3) (mainly, I didn't want to say that the component model _is_ the format, since (at least to me) the "model" is a way of thinking and the concrete artifacts like the binary format reflect that way of thinking, or something like that.)

component-model/src/design/why-component-model.md
catamorphism: Already fixed, see the current diff :)

component-model/src/design/why-component-model.md
vados-cosmonic: Yeah that's perfect -- I'm not sure the modern developer really uses `file` as much as in the past, but definitely nice to at least leave that for those who do/can!

component-model/src/design/why-component-model.md
vados-cosmonic: That looks good!

component-model/src/design/why-component-model.md
vados-cosmonic: So that's kind of true, but I think it's not unreasonable to think about it it as interacting via the component model itself. A component wraps modules, and sort of... intermediates their inputs and outputs -- so it's a bit murky. 

It's also *technically* possible to have a component that does *not* have an internal module actually. For example there are sometimes... patch/adapter subcomponents (not modules IIRC) that exist to patch imports to exports and do some other very internal things.

component-model/src/design/why-component-model.md
vados-cosmonic: Ah, so here, my thought was that the *contrast* is how you know it's a Component. Regular modules will output the other message (with "MVP") and the other code. 

Honestly one of the easiest ways most people do it is to run `wasm-tools print <file> | less` and check if `(component` is at the top...

component-model/src/design/why-component-model.md
vados-cosmonic: Yep, that's fine with me! I thought of "Component Model" as a proper noun here but it's not a big deal at all

component-model/src/design/why-component-model.md
vados-cosmonic: yeah that update certain works for me -- the wording definitely hits the spot now.

component-model/src/design/why-component-model.md
vados-cosmonic: OK I really want to get this in so let's add this in a follow up if we want to -- IMO it's worth adding but I could definitely be convinced otherwise

component-model/src/design/why-component-model.md
vados-cosmonic: Yeah this definitely works for me, the new prose does well to set up the why/landscape

Issue #279, opened 2025-07-13
Fixes & possible rename to "language support" section


vados-cosmonic:
While going through https://github.com/bytecodealliance/component-docs/pull/277, I noticed that the language support section could maybe be improved:

1. The name of the section does not *quite* reflect what it is 
2. The language agnostic tooling section should likely be it's own page

With regards to (1), it seems like we're listing how to create components in various languages, so maybe we should call it that? 

With regards to (2), it's weird that the language agnostic tooling section is *not* it's own page (and every other language is).

catamorphism:
I think #282 fixed this?

PR #282, opened 2025-07-16
Split "Language Agnostic Tooling" into its own page and fix example

catamorphism:
Per #279  I split off "Language Agnostic Tooling" into a separate page. 

Also, the example didn't work for me (with wasm-tools 1.235.0); the export in the .wat file was wrong and the function name, preceded by a '#', needed to be appended to the export name.

I plan to work more on this page, but I wanted to fix the example now so that no one gets led astray!


component-model/src/SUMMARY.md
```diff
@@ -15,7 +15,8 @@
 
 # Using WebAssembly Components
 
-- [Language Support for Components](./language-support.md)
+- [Creating Components in Specific Source Languages](./language-support.md)
```
vados-cosmonic: Hey @macovedj I know we just had a chat about this in another PR -- thoughts here?

component-model/src/language-support.md
```diff
@@ -1,9 +1,9 @@
...
+# Creating components in specific source languages
 
-WebAssembly can be targeted by the majority of top programming
-languages; however, the level of
-support varies. This document details the subset of languages that target WASI and support
-components.
+Many popular programming languages can be compiled to WebAssembly,
+but the level of support varies across languages.
+This document details languages with compilers and runtimes
+that support WebAssembly with WASI as a target language.
```
vados-cosmonic:
```suggestion
that support WebAssembly with WASI as a target platform.
```

Or just 'target' maybe?

component-model/src/SUMMARY.md
```diff
@@ -15,7 +15,8 @@
...
+  - [Language-Agnostic Tooling](./language-support/language-agnostic.md)
```
vados-cosmonic: NIT: Any reason to not go all the way to `language-agonistic-tooling.md` here?

component-model/src/language-support/language-agnostic.md
```diff
@@ -0,0 +1,62 @@
+## Language Agnostic Tooling
+
+### Building a Component with `wasm-tools`
```
vados-cosmonic:
```suggestion
### Building a Component from WebAssembly Text Format (WAT) with `wasm-tools`
```

component-model/src/language-support/language-agnostic.md
```diff
@@ -0,0 +1,62 @@
...
+working with WebAssembly modules and components.
+
+`wasm-tools` can be used to create a component from WAT.
+Here's how to create a component from WAT
+that implements the [`adder` world](https://github.com/bytecodealliance/component-docs/blob/main/component-model/examples/tutorial/wit/adder/world.wit)
+and simply adds two numbers.
+
+1. Install [`wasm-tools`](https://github.com/bytecodealliance/wasm-tools/tree/main#installation), a
+   tool for low-level manipulation of Wasm modules and components.
+2. The `add` function is defined inside the following world.
```
vados-cosmonic:
```suggestion

2. The `add` function is defined inside the following world.
```
Just making the spacing consistent here -- if this messes up the markdown feel free to ignore!

component-model/src/language-support/language-agnostic.md
```diff
@@ -0,0 +1,62 @@
...
+         local.get $rhs
+         i32.add)
+     (export "docs:adder/add@0.1.0#add" (func $add))
+   )
+   ```
+
+4. Use `wasm-tools` to create a binary core module with component metadata embedded inside it:
+
+   ```sh
+   $ wasm-tools component embed adder.wit add.wat -o add.wasm
```
vados-cosmonic:
```suggestion
   wasm-tools component embed adder.wit add.wat -o add.wasm
```

component-model/src/language-support/language-agnostic.md
```diff
@@ -0,0 +1,62 @@
...
+   ```
+
+5. Use `wasm-tools` to create a new component `.wasm` file
+   from the binary core module you just created:
+
+   ```sh
+   $ wasm-tools component new add.wasm -o add.component.wasm
```
vados-cosmonic:
```suggestion
   wasm-tools component new add.wasm -o add.component.wasm
```

component-model/src/language-support/language-agnostic.md
```diff
@@ -0,0 +1,62 @@
...
+that implements the [`adder` world](https://github.com/bytecodealliance/component-docs/blob/main/component-model/examples/tutorial/wit/adder/world.wit)
+and simply adds two numbers.
+
+1. Install [`wasm-tools`](https://github.com/bytecodealliance/wasm-tools/tree/main#installation), a
+   tool for low-level manipulation of Wasm modules and components.
+2. The `add` function is defined inside the following world.
+   Create a file called `adder.wit` whose contents are as follows:
+
+   ```wit
+   package docs:adder@0.1.0;
```
vados-cosmonic: What do you think about using the markdown inclusion functionality here -- it's something I constantly forget to do, but we have this file in the repo IIRC

component-model/src/language-support/language-agnostic.md
```diff
@@ -0,0 +1,62 @@
...
+   $ wasm-tools component embed adder.wit add.wat -o add.wasm
+   ```
+
+5. Use `wasm-tools` to create a new component `.wasm` file
+   from the binary core module you just created:
+
+   ```sh
+   $ wasm-tools component new add.wasm -o add.component.wasm
+   ```
+
```
vados-cosmonic: Do you think it's worth noting that `.component.wasm` is a convention and not a hard-and-fast rule or some sort of recognized suffix here?

component-model/src/language-support/language-agnostic.md
```diff
@@ -0,0 +1,62 @@
...
+### Running a Component with Wasmtime
+
+You can "run" a component by calling one of its exports. Hosts and runtimes often only support
+running components with certain exports. The [`wasmtime`](https://github.com/bytecodealliance/wasmtime) CLI can only run "command" components, so in
+order to run the `add` function above, it first must be composed with a primary "command" component
+that calls it. See [documentation on running components](../running-components/wasmtime.md) for more details.
```
vados-cosmonic: So a few thoughts here --

- Should this section even be in this file (seems like we're jumping forward quite a bit, but I understand the urge to show the usability of what we just created)
- Now that `wasmtime run --invoke` exists, we can actually run the export quite easily
- I wonder if we should be more explicit about the fact that any compliant webassembly runtime that supports components can run the component.
- If we keep this section, should we post the output here?

component-model/src/language-support/language-agnostic.md
```diff
@@ -0,0 +1,62 @@
+## Language Agnostic Tooling
+
+### Building a Component with `wasm-tools`
+
+[`wasm-tools`](https://github.com/bytecodealliance/wasm-tools) provides a suite of subcommands for
```
vados-cosmonic: Should we introduce WAT first *then* introduce `wasm-tools`?

Also, I wonder if we should mention [`wat2wasm`](https://github.com/WebAssembly/wabt) and note that it doesn't create components?

component-model/src/language-support.md
catamorphism: Using "target" as a verb feels a little more jargon-y to me. Just vibes, though.

component-model/src/SUMMARY.md
catamorphism: Do you mean renaming `language-agnostic.md` to `language-agnostic-tooling.md`? That would be fine, no strong feelings one way or the other.

component-model/src/language-support.md
vados-cosmonic: yeah I could do with out it -- what do you think about just "platform"?

component-model/src/SUMMARY.md
vados-cosmonic: Yup, exactly!

component-model/src/language-support/language-agnostic.md
catamorphism: That would be great (the file is in the repo), but I couldn't find a way to include files using pure Markdown. I found mdbook-utils, which supports an `{{# include file}}` [syntax](https://john-cd.com/mdbook-utils/definitions/code_blocks_and_includes.html), but that's not part of the mdbook install that's documented in the repo. ~~So I'll file a separate issue for this.~~ Filed #283

vados-cosmonic: Ah, so we do [this](https://github.com/bytecodealliance/component-docs/blob/e516c563df5b993b93676a00401fe64c0815568f/component-model/src/language-support/rust.md?plain=1#L69)

catamorphism: Oh, great. Fixed in [263139e](https://github.com/bytecodealliance/component-docs/pull/282/commits/263139ea0dcb2efbae069a2e7b451388684ab2d8)

component-model/src/language-support/language-agnostic.md
catamorphism: Added in [a440d3f](https://github.com/bytecodealliance/component-docs/pull/282/commits/a440d3f55429da335091c7493640d1ba73c8fa6a)

component-model/src/language-support.md
catamorphism: "Platform" is good, I integrated that suggestion!

component-model/src/language-support/language-agnostic.md
catamorphism: I think the section should be there; otherwise it's pretty unsatisfying to not know how to run the code you just created. 

I updated it to use `wasmtime run --invoke` and removed some of the text that's now superfluous, in [95cb5dc](https://github.com/bytecodealliance/component-docs/pull/282/commits/95cb5dcd5a5c1d4f4dab74f5b0fb7ed6d0a92656).

vados-cosmonic: In that case maybe it makes sense to actually move this to a section that is under Language support normally? It may not be a large stretch to think of WAT as a language that compiles to WebAssembly! "WebAssembly Text Format (WAT)" might make a good section heading. 

I originally thought this section would be more about how to do things with `wasm-tools` *other* than simply creating a component with WAT, but if that's the case, maybe we should be treating WAT more similarly to the other languages

catamorphism: I can see where that makes sense for someone already familiar with Wasm, but from the perspective of someone who's not very familiar with it, I think the current structure makes sense. If a beginner thinks of languages that might compile to Wasm, they might think, "C++, Rust, JavaScript..." but they wouldn't think of WAT. I could be persuaded otherwise, though.

component-model/src/language-support/language-agnostic.md
catamorphism: In [49bdac0](https://github.com/bytecodealliance/component-docs/pull/282/commits/49bdac0bf0514622c480257865b0cbb07d855659) I added a section introducing WAT and walking through the example WAT code. Let me know if you think it's too much detail (or not enough!)

component-model/src/SUMMARY.md
catamorphism: I'll do this now. Hopefully renaming the file doesn't make the diffs too hard to follow (on the other hand, maybe it's changed enough that the diffs aren't useful anyway...)

catamorphism: Done in [09ab254](https://github.com/bytecodealliance/component-docs/pull/282/commits/09ab254b9bcf9c18ee13585d33d91718023d8cb0)

component-model/src/language-support.md
```diff
@@ -20,10 +20,10 @@ components within and among toolchains.
 Each section covers how to build and
 run components for a given toolchain:
 
-- [Wasm Language Support](#wasm-language-support)
-  - [Language Agnostic Tooling](#language-agnostic-tooling)
-    - [Building a Component with `wasm-tools`](#building-a-component-with-wasm-tools)
-    - [Running a Component with Wasmtime](#running-a-component-with-wasmtime)
+  - [Language Agnostic Tooling](./language-support/language-agnostic-tooling.md)
+    - [WAT (WebAssembly Text Format)](./language-support/language-agnostic-tooling.md#wat-webassembly-text-format)
+    - [Building a Component from WAT with `wasm-tools`](./language-support/language-agnostic-tooling.md#building-a-component-with-wasm-tools)
```
itowlson: This seems extremely "advanced topics" to me.  It's very niche in its use cases, and challenging to get right for all but the most trivial examples.  (Or indeed, given the error Tim found in the adder, _even_ the most trivial examples!)  I would keep this out of the main flow of the book; I would certainly not make readers scroll past it to get the the "how to run a component in Wasmtime" bit.

(I appreciate some of this already exists but it seems to be getting more in-your-face.  Perhaps I'm misreading.)

component-model/src/SUMMARY.md
```diff
@@ -15,7 +15,8 @@
 
 # Using WebAssembly Components
 
-- [Language Support for Components](./language-support.md)
+- [Creating Components in Specific Source Languages](./language-support.md)
+  - [Language-Agnostic Tooling](./language-support/language-agnostic-tooling.md)
```
itowlson: My brain graunches its gears when the section is titled "specific source languages" and the first thing in it is "language agnostic"!

component-model/src/language-support.md
catamorphism: That's fair, maybe add it to the "Advanced Topics" section before "Canonical ABI"?

component-model/src/SUMMARY.md
catamorphism: Good catch :) But I'll wait to change it until @vados-cosmonic weighs in on the possibility of moving the section to "advanced topics".

component-model/src/language-support/language-agnostic.md
itowlson: WARNING! HOT TAKE!  Folks close to Wasm vastly overestimate the interest and value of WAT to users.  I realise it's great for Wasm implementers, like when you're building a feature and it's not yet supported by major language compilers.  But for Wasm _users_, who were the original target audience of the book, it's as relevant as knowing Java bytecode.

So if we're reworking this section of the book then I'd argue to _reduce_ the amount and prominence of WAT content in the main flow, not to increase it.

If other folks have different experience, then of course I defer to that, but in 4 or 5 years, I've only busted out WAT on a handful of occasions; and except maybe once that was to read it in a debugging context, not to write it...!

t3hmrman: > WARNING! HOT TAKE! Folks close to Wasm vastly overestimate the interest and value of WAT to users. I realise it's great for Wasm implementers, like when you're building a feature and it's not yet supported by major language compilers. But for Wasm users, who were the original target audience of the book, it's as relevant as knowing Java bytecode.

This is absolutely correct. WAT is wonderful/extremely useful, but just like most people don't actually want to write Assembly (or Lisp!) most people don't actually get super into WAT, I think.

vados-cosmonic: This is absolutely correct. WAT is wonderful/extremely useful, but just like most people don't actually want to write Assembly (or Lisp!) most people don't actually get super into WAT, I think.

t3hmrman: I think my thinking here stays the same -- WAT is a language that compiles to WebAssembly, and I think we should treat it as such (and maybe even put it at the bottom of the listing). Most people who are familiar with a language already on the list will jump there, and at this point "language independent tooling" is a bit of a misnomer.

Just like assembly is a language that one can write code in, so is WAT

component-model/src/language-support.md
vados-cosmonic: So I think my thinking here is that WAT is a language -- and I fully agree with @itowlson here that it is niche. Most people will not write it (a point I think we emphasize later IIRC -- most people will use some other programming language).

The sections we *want* to include with WAT make sense -- we do them for the other languages, so I think treating WAT as a language accomplishes *both* goals simultaneously; showing how to build a basic component with WAT and how to run it.

So to restate, I think WAT makes sense as a language section, just like the others (it is fine to have it less specified at first, it doesn't have to be as involved IMO).

component-model/src/SUMMARY.md
vados-cosmonic: Ah so yeah I don't know how @itowlson feels here about keeping it in the same section but treating it as a language and maybe moving it towards the bottom of the list (or alphabetical? which might be the bottom anyway) for consistency and de-prioritizing it.

component-model/src/SUMMARY.md
vados-cosmonic: An update here since we discussed this on the docs call, this seems fine!

component-model/src/SUMMARY.md
vados-cosmonic: Oh also, with regards to "Specific Source Languages" yeah I don't think we need that -- I think "Creating Components" works just fine -- it's abundantly clear and I think we can expect people to gather that the listing is programming languages.

catamorphism: In [7834e8c](https://github.com/bytecodealliance/component-docs/pull/282/commits/7834e8c68d30f98a0d0840ec5ef190492266695b) I've changed "creating components in specific source languages" to just "creating components". I also moved WAT last in the list and renamed the file from `language-agnostic-tooling.md` to `wat.md`.

component-model/src/language-support.md
catamorphism: I moved it to the end of the list of languages in [7834e8c](https://github.com/bytecodealliance/component-docs/pull/282/commits/7834e8c68d30f98a0d0840ec5ef190492266695b).

component-model/src/language-support/language-agnostic.md
catamorphism: I think I've resolved this by moving it to the end of the list of languages? (See [7834e8c](https://github.com/bytecodealliance/component-docs/pull/282/commits/7834e8c68d30f98a0d0840ec5ef190492266695b))

component-model/src/SUMMARY.md
vados-cosmonic: Thanks! that certainly looks good to me!

component-model/src/language-support/language-agnostic.md
vados-cosmonic: Yup! thank you -- this got brought up a bunch of places...

component-model/src/language-support.md
```diff
@@ -1,9 +1,9 @@
-# Wasm Language Support
+# Creating components in specific source languages
```
vados-cosmonic:
```suggestion
# Creating components
```

vados-cosmonic (approved):
LGTM :rocket: 

Thanks for the awesome work here @catamorphism!

PR #284, opened 2025-07-21
Remove wasm-tools warning for Golang

alexcasalboni:
This has been fixed in https://github.com/bytecodealliance/go-modules/pull/298

Also see the discussion in https://github.com/bytecodealliance/go-modules/issues/297

I've re-tested today and it's working fine with the latest too (`wasm-tools@1.235.0`) 👌

### DIFF
--- a/component-model/src/language-support/go.md
+++ b/component-model/src/language-support/go.md
@@ -13,12 +13,6 @@ Follow the [TinyGo installation instructions](https://tinygo.org/getting-started
 
 Additionally, install the `wasm-tools` CLI tool from the [wasm-tools repository](https://github.com/bytecodealliance/wasm-tools/releases).
 
-> [!WARNING]
-> Due to some upstream issues, only `wasm-tools` versions 1.225.0 or earlier can be used with `wit-bindgen-go`
->
-> If using the Rust toolchain to install `wasm-tools`, it can be installed like so:
-> `cargo install --locked wasm-tools@1.225.0 --force`
-
 To verify the installation, run the following commands:
 
 ```
--- a/component-model/src/language-support/go.md
+++ b/component-model/src/language-support/go.md
@@ -13,6 +13,16 @@ Follow the [TinyGo installation instructions](https://tinygo.org/getting-started
 
 Additionally, install the `wasm-tools` CLI tool from the [wasm-tools repository](https://github.com/bytecodealliance/wasm-tools/releases).
 
+> [!WARNING]
+> `wit-bindgen-go` comes with its own `wasm-tools` vendored version, but tinygo still requires you to install it.
+> Even if unlikely, this could lead to version mismatch when using older versions of `wasm-tools`.
+> Please make sure to keep your local `wasm-tools` udpated, should you encounter any issues.
+>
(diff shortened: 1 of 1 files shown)


vados-cosmonic:
Hey @alexcasalboni thanks for making this PR -- could we add something that tells people which version this *works* with? It would be useful to know the lower bound on the version they should be using until we're well clear of the broken versions.

alexcasalboni:
@vados-cosmonic afaik, the issue this warning mentions was just a temporary issue with `wit-bindgen-go` and its own dependencies.

This is fixed since `wit-bindgen-go@v0.6.2`.

vados-cosmonic:
Yep, that's right -- the fix was made in `wit-bindgen-go` right? The situation I want to avoid is someone with an older version of `wit-bindgen-go` running into this issue and being confused!

A line like "Please ensure you're using version wit-bindgen-go v0.6.2 or above" *should* make it easy to avoid that situation. If this situation is incredibly unlikely then we can skip it though.

alexcasalboni:
Yes correct, the fix was made in `wit-bindgen-go`, back in March.

I was looking for a way to pinpoint the `wit-bindgen-go` version, but this Go guide is now based on `Go 1.24` and suggests using the new `go tool` approach, which doesn't support specifying a tool version.

So my understanding is that we're ok with saying "Automatically use the latest `wit-bindgen-go` version and that should be compatible with all `wasm-tools` versions" 😄  If that's not the case, we should switch back to using a regular dependency (not an unversioned tool).

Just in case, I've done some additional tests to confirm that the latest version of `wit-bindgen-go` works fine with all `wasm-tools` versions (even the old `1.225.0`, `1.226.0`, `1.227.0`, etc.). And surprisingly, you still need to install it yourself (even if the [v0.6.0](https://github.com/bytecodealliance/go-modules/releases/tag/v0.6.0) release mentions that "wasm-tools is now vendored as a WebAssembly module, executed using Wazero."). So I'm a bit confused 🤔 

@ydnar could you please share your thoughts here? I think I am missing something important 🙏

vados-cosmonic:
Hmmm OK, so we'd probably want to go through the guide with newer versions of all the tools and make sure things still work smoothly... 

Along with that, we'd probably want to include a warning that points to versioning being somewhat touchy -- more to encourage people to make sure they have the exact versions (or newer) that we're using in the guide.

How does that sound?

alexcasalboni:
Agreed 👍 

We could keep a warning like this:

```markdown
> [!WARNING]
> `wit-bindgen-go` comes with its own `wasm-tools` vendored version, but tinygo still requires you to install it.
> Even if unlikely, this could lead to version mismatch when using older versions of `wasm-tools`.
> Please make sure to keep your local `wasm-tools` udpated, should you encounter any issues.
>
> If using the Rust toolchain to install `wasm-tools`, it can be installed like so:
> `cargo install --locked wasm-tools@1.235.0 --force`
> or via cargo binstall:
> `cargo binstall wasm-tools@1.235.0`
```

(please feel free to rephase)

vados-cosmonic (approved):
Thanks for this update, @alexcasalboni !

PR #287, opened 2025-07-21
Add 'WIT by Example' section

catamorphism:
I thought it might be helpful to add some WIT examples before launching into the detailed overview.


vados-cosmonic:
Hey @catamorphism thanks for adding this section -- generally we like to discuss additions like this in issues *first* before putting up PRs about them and ideally maybe even in a meeting beforehand.

I think these changes are great, but would love to socialize the changes more broadly before reviewing/making a move here. @itowlson @kate-goldenring @macovedj thoughts? 

It would be great to walk through this at the next meeting if this doesn't merge before then.

vados-cosmonic:
> As a reviewer - and I feel a bit silly bringing it up - I'm finding it increasingly hard to follow the raw text/diff view layout. There's places where we use 40-character lines and places where we use full-length lines, and either is fine but in the context of full-length lines the sections of short lines give me an extremely staccato effect. I appreciate that line lengths are never going to be completely even and that nobody wants to spend their life reflowing text as they edit, and I'm not saying we should have rigid guidelines - we shouldn't - but I'd find it way easier to read if the paragraphs flowed more like paragraphs.

Concretely with regards to this, we should probably standardize the line wrap we expect. I don't want to be restrictive about it, but at least a posted policy would be good (and if we could make it an automated lint that would be perfect, because people don't get mad at robots for being pedantic).

itowlson:
> people don't get mad at robots for being pedantic

I assure you I get mad at formatter lints _all the time._

More seriously, I'd favour loose policy over pedantic automation.  I don't want my PR rejected because a line strayed up to 81 characters or down to 49.  I'd rather trust loose consensus and human good taste, even if humans differ!

vados-cosmonic:
Hey just checking in here but @catamorphism I know we discussed this at the last meeting -- you're not blocked on us here right? IIRC we had a concrete list of things that would be nice to add to get this across the line.

component-model/examples/wit-section-examples/clocks/monotonic-clock.wit
```diff
@@ -0,0 +1,29 @@
+package wasi:clocks@0.2.6;
```
pchickey: Can these examples use a package namespace other than `wasi`? perhaps `wasi-example`?

vados-cosmonic: A little comment identifying that these examples are small facsimiles of the "real" WASI interfaces might also be helpful (maybe with a link to the real repo) so people can explore on their own?

vados-cosmonic:
Hey @catamorphism I think we're OK to merge this as-is once it's ready!

component-model/examples/wit-section-examples/clocks/monotonic-clock.wit
catamorphism: Done in [1183a96](https://github.com/bytecodealliance/component-docs/pull/287/commits/1183a9638c9c89dfd92dcc77c88cc43d9a99c9d9) and [605a2e1](https://github.com/bytecodealliance/component-docs/pull/287/commits/605a2e1b731b1a2b84b58b913fb62fd486c68df0).

PR #288, opened 2025-07-21
Expand WIT overview section

catamorphism:
Most of the changes here are to add semantic line breaks, but I did expand on some concepts, particularly resources.

### DIFF
--- a/component-model/src/design/wit.md
+++ b/component-model/src/design/wit.md
@@ -356,7 +356,7 @@ flags allowed-methods {
 ### Type aliases
 
 You can define a new type alias using `type ... = ...`.
-This can be useful for giving shorter or more meaningful names to types:
+Type aliases are useful for giving shorter or more meaningful names to types:
 
 ```wit
 type buffer = list<u8>;
--- a/component-model/src/design/wit.md
+++ b/component-model/src/design/wit.md
@@ -1,6 +1,17 @@
 # An Overview of WIT
 
-The WIT (Wasm Interface Type) language is used to define Component Model [interfaces](#interfaces) and [worlds](#worlds). WIT isn't a general-purpose coding language and doesn't define behaviour; it defines only _contracts_ between components. This topic provides an overview of key elements of the WIT language. The official WIT specification and history can be found in the [`WebAssembly/component-model` repository](https://github.com/WebAssembly/component-model/blob/main/design/mvp/WIT.md).
+We've explained how worlds and interfaces
+define how components relate to each other.
+To define a new component, you will need to define worlds and interfaces
+by writing code in the Wasm Interface Type (WIT) language.
+WIT also serves as documentation for existing components
+that you may wish to use.
+
+The WIT (Wasm Interface Type) language is used to define Component Model [interfaces](#interfaces) and [worlds](#worlds).
+WIT isn't a general-purpose programming language and doesn't define behaviour;
+it defines only _contracts_ between components.
+This topic provides an overview of key elements of the WIT language.
+The official WIT specification and history can be found in the [`WebAssembly/component-model` repository](https://github.com/WebAssembly/component-model/blob/main/design/mvp/WIT.md).
 
 - [An Overview of WIT](#an-overview-of-wit)
   - [Structure of a WIT file](#structure-of-a-wit-file)
@@ -66,14 +77,19 @@ print-hello: func();
 
 ## Identifiers
 
-WIT identifiers have a slightly different set of rules from what you might be familiar with from, say, C, Rust, or Java.  These rules apply to all names - types, functions, interfaces, and worlds. (Package identifiers are a little more complex and will be covered in the [Packages section](#packages).)
+_Identifiers_ are names for variables, functions, types, interfaces, and worlds.
+WIT identifiers have a slightly different set of rules
+from what you might be familiar with in languages like C, Rust, and Java.
+These rules apply to all names, except for packages.
+Package identifiers are a little more complex and will be covered in the [Packages section](#packages).
 
-* Identifiers are restricted to ASCII `kebab-case` - sequences of words, separated by single hyphens.
+* Identifiers are restricted to ASCII `kebab-case`: sequences of words, separated by single hyphens.
   * Double hyphens (`--`) are not allowed.
   * Hyphens aren't allowed at the beginning or end of the sequence, only between words.
 * An identifier may be preceded by a single `%` sign.
-  * This is _required_ if the identifier would otherwise be a WIT keyword. For example, `interface` is **not** a legal identifier, but `%interface` is legal.
-* Each word in the sequence must begin with an ASCII letter, and may contain only ASCII letters and digits.  
+  * This is _required_ if the identifier would otherwise be a WIT keyword.
+    For example, `interface` is **not** a legal identifier, but `%interface` is legal.
+* Each word in the sequence must begin with an ASCII letter, and may contain only ASCII letters and digits.
   * A word cannot begin with a digit.
   * A word cannot contain a non-ASCII Unicode character.
   * A word cannot contain punctuation, underscores, etc.
@@ -96,13 +112,18 @@ WIT defines the following primitive types:
 | `u8`, `u16`, `u32`, `u64`  | Unsigned integers of the appropriate width. For example, `u32` is an unsigned 32-bit integer. |
 | `f32`, `f64`               | Floating-point numbers of the appropriate width. For example, `f64` is a 64-bit (double precision) floating-point number. See the note on `NaN`s below. |
 | `char`                     | Unicode character. (Specifically, a [Unicode scalar value](https://unicode.org/glossary/#unicode_scalar_value).) |
-| `string`                   | A Unicode string - that is, a finite sequence of characters. |
+| `string`                   | A Unicode string: that is, a finite sequence of characters. |
 
-> The `f32` and `f64` types support the usual set of IEEE 754 single and double-precision values, except that they logically only have a single `nan` value. The exact bit-level representation of an IEEE 754 `NaN` is not guaranteed to be preserved when values pass through WIT interfaces as the singular WIT `nan` value.
+> The `f32` and `f64` types support the usual set of IEEE 754 single and double-precision values,
+> except that they logically only have a single `nan` value.
+> The exact bit-level representation of an IEEE 754 `NaN`
+> is not guaranteed to be preserved when values pass through WIT interfaces
+> as the singular WIT `nan` value.
 
 ### Lists
 
-`list<T>` for any type `T` denotes an ordered sequence of values of type `T`.  `T` can be any type, built-in or user-defined:
+`list<T>` for any type `T` denotes an ordered sequence of values of type `T`.
+`T` can be any type, built-in or user-defined:
 
 ```wit
 list<u8>       // byte buffer
@@ -113,7 +134,10 @@ This is similar to Rust `Vec`, or Java `List`.
 
 ### Options
 
-`option<T>` for any type `T` may contain a value of type `T`, or may contain no value.  `T` can be any type, built-in or user-defined.  For example, a lookup function might return an option, allowing for the possibility that the lookup key wasn't found:
+`option<T>` for any type `T` may contain a value of type `T`, or may contain no value.
+`T` can be any type, built-in or user-defined.
+For example, a lookup function might return an option in order to allow
+for the possibility that the lookup key wasn't found:
 
 ```wit
 option<customer>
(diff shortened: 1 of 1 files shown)


component-model/src/design/wit.md
```diff
@@ -113,39 +134,61 @@ This is similar to Rust `Vec`, or Java `List`.
...
-> This is a special case of a [variant](#variants) type.  WIT defines it so that there is a common way of expressing it, so that you don't need to create a variant type for every value type, and to enable it to be mapped idiomatically into languages with option types.
+> This is a special case of a [variant](#variants) type.
+> WIT defines it so that there is a common way of expressing it,
+> so that you don't need to create a variant type for every value type,
+> and to enable it to be mapped idiomatically into languages with option types.
 
 ### Results
 
-`result<T, E>` for any types `T` and `E `may contain a value of type `T` _or_ a value of type `E` (but not both). This is typically used for "value or error" situations; for example, a HTTP request function might return a result, with the success case (the `T` type) representing a HTTP response, and the error case (the `E` type) representing the various kinds of error that might occur:
+`result<T, E>` for any types `T` and `E
```
vados-cosmonic:
```suggestion
`result<T, E>` for any types `T` and `E`
```

component-model/src/design/wit.md
```diff
@@ -113,39 +134,61 @@ This is similar to Rust `Vec`, or Java `List`.
...
+may contain a value of type `T` _or_ a value of type `E`
+(but not both).
+This is typically used for "value or error" situations:
+for example, a HTTP request function might return a result,
```
vados-cosmonic: This kind of becomes a run-on, I wonder if we can just delete the line about typical use and start from "for example"

component-model/src/design/wit.md
```diff
@@ -113,39 +134,61 @@ This is similar to Rust `Vec`, or Java `List`.
...
 
 ```wit
 result<http-response, http-error>
 ```
 
 This is similar to Rust `Result`, or Haskell `Either`.
 
-> This is a special case of a [variant](#variants) type.  WIT defines it so that there is a common way of expressing it, so that you don't need to create a variant type for every combination of value and error types, and to enable it to be mapped idiomatically into languages with result or "either" types.
+> This is a special case of a [variant](#variants) type.
+> WIT defines it so that there is a common way of expressing it,
```
vados-cosmonic:
```suggestion
> WIT defines the `result` type so that there is a common way of expressing this behavior,
```

Unrelated to these changes but would be nice to improve this while we're here

component-model/src/design/wit.md
```diff
@@ -113,39 +134,61 @@ This is similar to Rust `Vec`, or Java `List`.
...
+> so that you don't need to create a variant type for every combination of value and error types,
```
vados-cosmonic:
```suggestion
> so that developers don't need to create variant types for every combination of value and error types,
```

component-model/src/design/wit.md
```diff
@@ -113,39 +134,61 @@ This is similar to Rust `Vec`, or Java `List`.
...
+but has nothing to return if it succeeds.
+In this case, you can omit the corresponding type as follows:
 
 ```wit
 result<u32>     // no data associated with the error case
 result<_, u32>  // no data associated with the success case
 result          // no data associated with either case
 ```
 
+The underscore `_` stands in for "no data".
```
vados-cosmonic:
```suggestion
The underscore `_` stands in "no data" and is generally represented as the unit type in a target language (e.g. `()` in rust, `null` in Javsacript).
```

component-model/src/design/wit.md
```diff
@@ -156,11 +199,16 @@ This is similar to tuples in Rust or OCaml.
 
 ## User-defined types
 
-You can define your own types within an `interface` or `world`. WIT offers several ways of defining new types.
+You can define your own types within an `interface` or `world`.
+WIT offers several ways of defining new types.
```
vados-cosmonic: Could we collapse this into something like

"New domain specific types can be defined within an `interface` or `world`"

component-model/src/design/wit.md
```diff
@@ -203,20 +265,30 @@ enum color {
 }
 ```
 
-This can provide a simpler representation in languages without discriminated unions. For example, a WIT `enum` can translate directly to a C++ `enum`.
+This can provide a simpler representation in languages without discriminated unions.
+For example, a WIT `enum` can translate directly to a C `enum`.
```
vados-cosmonic:
```suggestion
For example, a WIT `enum` can translate directly to a C/C++ `enum`.
```

I assume the ++ drop was intentional but maybe we can include both here?

component-model/src/design/wit.md
```diff
@@ -203,20 +265,30 @@ enum color {
...
 
 ### Resources
 
-Resources are handles to some entity that lives outside of the component. They
-describe things that can't or shouldn't be copied by value; instead, their
-ownership or reference can be passed between two components via a handle. Unlike
-other WIT types which are simply plain data, resources only expose behavior
-through methods. Resources can be thought of as _objects that implement_ an
-interface.
+A resources is a handle to some entity that exists outside of the component.
```
vados-cosmonic:
```suggestion
A resource is a handle to some entity that exists outside of the component.
```

component-model/src/design/wit.md
```diff
@@ -203,20 +265,30 @@ enum color {
...
+Resources describe things that shouldn't be passed by value
+because they can't or shouldn't be copied.
+Instead, they should be passed by reference.
```
vados-cosmonic:
```suggestion
Resources describe entities that can't or shouldn't be copied; entities that should
be passed by reference rather than by value.
```

What do you think about this summarization? Does it make it a bit more clear/concise?

component-model/src/design/wit.md
```diff
@@ -203,20 +265,30 @@ enum color {
...
+Components can pass resources to each other via a handle.
+They can pass ownership of resources, or pass non-owned references to resources.
+
+> If you're not familiar with the concepts of borrowing and ownership
+> for pointers, see [the Rust documentation](https://doc.rust-lang.org/book/ch04-01-what-is-ownership.html).
+
+Unlike other WIT types, which are simply plain data,
+resources only expose behavior through methods.
+Resources can be thought of as _objects that implement_ an interface.
+("Interface" here is used in the object-oriented programming sense,
```
vados-cosmonic, line 283: This is a very well-timed clarification

component-model/src/design/wit.md
```diff
@@ -229,35 +301,47 @@ resource blob {
 
 As shown in the `blob` example, a resource can contain:
 
-- _methods_: functions that implicitly take a `self` (often called `this` in many languages)
-parameter that is a handle
+- _methods_: functions that implicitly take a `self` parameter that is a handle.
```
vados-cosmonic:
```suggestion
- _methods_: functions that implicitly take a `self` (AKA `this`) parameter that is a handle.
```

Would this work to collapse the second parenthesized sentence here?

component-model/src/design/wit.md
```diff
@@ -229,35 +301,47 @@ resource blob {
...
 
-> More precisely, these are borrowed or owned `handles` of the resource. Learn more about `handles` in the [upstream component model specification](https://github.com/WebAssembly/component-model/blob/main/design/mvp/WIT.md#handles).
+> More precisely, these are borrowed or owned `handles` of the resource.
+> Learn more about `handles` in the [upstream component model specification](https://github.com/WebAssembly/component-model/blob/main/design/mvp/WIT.md#handles).
 
 ### Flags
 
-A `flags` type is a set of named booleans.  In an instance of the type, each flag will be either `true` or `false`.
+A `flags` type is a set of named booleans.
+In an instance of the type, each flag will be either `true` or `false`.
```
vados-cosmonic:
```suggestion
```

We could probably remove this

component-model/src/design/wit.md
```diff
@@ -268,11 +352,13 @@ flags allowed-methods {
 }
 ```
 
-> A `flags` type is logically equivalent to a record type where each field is of type `bool`, but it is represented more efficiently (as a bitfield) at the binary level.
+> A `flags` type is logically equivalent to a record type where each field is of type `bool`,
+but it is represented more efficiently (as a bitfield) at the binary level.
```
vados-cosmonic:
```suggestion
> but it is represented more efficiently (as a bitfield) at the binary level.
```

component-model/src/design/wit.md
```diff
@@ -268,11 +352,13 @@ flags allowed-methods {
...
 
 ### Type aliases
 
-You can define a new named type using `type ... = ...`. This can be useful for giving shorter or more meaningful names to types:
+You can define a new named type using `type ... = ...`.
```
vados-cosmonic:
```suggestion
You can define a new type alias using `type ... = ...`.
```

component-model/src/design/wit.md
```diff
@@ -268,11 +352,13 @@ flags allowed-methods {
...
+This can be useful for giving shorter or more meaningful names to types:
```
vados-cosmonic:
```suggestion
Type aliases are useful for giving shorter or more meaningful names to types:
```

component-model/src/design/wit.md
```diff
@@ -298,17 +387,20 @@ add: func(a: u64, b: u64) -> u64;
 lookup: func(store: kv-store, key: string) -> option<string>;
 ```
 
-For multi-value returns from a function, you can use a [tuple](#tuple) or [record](#record).
+To express a function that returns multiple values,
+you can use a [tuple](#tuple) or [record](#record).
```
vados-cosmonic:
```suggestion
you can use any compound type ([tuples](#tuple), [record](#record), etc).
```

component-model/src/design/wit.md
```diff
@@ -345,13 +440,23 @@ interface canvas {
 }
 ```
 
-> Even if you are only using one type, it must still be enclosed in braces. For example, `use types.{dimension}` is legal but `use types.dimension` is not.
+The `canvas` interface uses the types `dimension` and `point`
+declared in the `types` interface.
```
vados-cosmonic:
```suggestion
The `canvas` interface uses the types `dimension` and `point` declared in the `types` interface.
```

This would fit on one line right?

component-model/src/design/wit.md
```diff
@@ -385,9 +501,13 @@ world http-proxy {
...
 
-> As this example shows, import and export apply at the interface level, not the package level. You can import one interface defined in a package, while exporting another interface defined in the same package. Packages group definitions; they don't represent behaviour.
+> As this example shows, import and export apply at the interface level, not the package level.
+> You can import one interface defined in a package,
+> while exporting another interface defined in the same package.
+> A package groups definitions together; it doesn't describe a coherent set of behaviours.
 
-WIT does not define how packages are resolved - different tools may resolve them in different ways.
+WIT does not define how packages are resolved;
+different tools may resolve them in different ways.
```
vados-cosmonic:
```suggestion
WIT does not define how packages are resolved; different tools may resolve them in different ways.
```

vados-cosmonic (commented):
Thanks for expanding this section! Left a bunch of comments, mostly nits

component-model/src/design/wit.md
catamorphism: Done in [929bd2b](https://github.com/bytecodealliance/component-docs/pull/288/commits/929bd2b88402716e869aceb137aec8d274a2b292)

component-model/src/design/wit.md
catamorphism: Done in [929bd2b](https://github.com/bytecodealliance/component-docs/pull/288/commits/929bd2b88402716e869aceb137aec8d274a2b292)

component-model/src/design/wit.md
```diff
@@ -1,6 +1,17 @@
 # An Overview of WIT
 
-The WIT (Wasm Interface Type) language is used to define Component Model [interfaces](#interfaces) and [worlds](#worlds). WIT isn't a general-purpose coding language and doesn't define behaviour; it defines only _contracts_ between components. This topic provides an overview of key elements of the WIT language. The official WIT specification and history can be found in the [`WebAssembly/component-model` repository](https://github.com/WebAssembly/component-model/blob/main/design/mvp/WIT.md).
+We've explained how worlds and interfaces
+define how components relate to each other.
```
vados-cosmonic: Probably can remove this -- someone may land on this page without going to the other ones, I think. 

If we want to refer to the previous section, then we maybe could include a link and a statement that is more general/doesn't expect the transition from the previous page.

component-model/src/design/wit.md
```diff
@@ -1,6 +1,17 @@
...
+To define a new component, you will need to define worlds and interfaces
+by writing code in the Wasm Interface Type (WIT) language.
+WIT also serves as documentation for existing components
+that you may wish to use.
+
+The WIT (Wasm Interface Type) language is used to define Component Model [interfaces](#interfaces) and [worlds](#worlds).
+WIT isn't a general-purpose programming language and doesn't define behaviour;
+it defines only _contracts_ between components.
```
vados-cosmonic: These two lines feel like they should be the first ones

component-model/src/design/wit.md
```diff
@@ -203,20 +264,29 @@ enum color {
...
-through methods. Resources can be thought of as _objects that implement_ an
-interface.
+A resource is a handle to some entity that exists outside of the component.
+Resources describe entities that can't or shouldn't be copied:
+entities that should be passed by reference rather than by value.
+Components can pass resources to each other via a handle.
+They can pass ownership of resources, or pass non-owned references to resources.
+
+> If you're not familiar with the concepts of borrowing and ownership
+> for pointers, see [the Rust documentation](https://doc.rust-lang.org/book/ch04-01-what-is-ownership.html).
```
vados-cosmonic:
```suggestion
> for references, see [the Rust documentation](https://doc.rust-lang.org/book/ch04-01-what-is-ownership.html).
```

component-model/src/design/wit.md
catamorphism: Changed in [81c91ef](https://github.com/bytecodealliance/component-docs/pull/288/commits/81c91efbc6166edd1370b0eb8e61a809f98cc8f0)

component-model/src/design/wit.md
catamorphism: Changed in [81c91ef](https://github.com/bytecodealliance/component-docs/pull/288/commits/81c91efbc6166edd1370b0eb8e61a809f98cc8f0)

vados-cosmonic (approved):
LGTM 🚀

PR #289, opened 2025-07-23
feat(lang/c): use wasm32-wasip2-clang binary directly

### LINKED ISSUE #276: Update the wasi c examples to use the wasi 0.2 tooling
The wasi [C example ](https://component-model.bytecodealliance.org/language-support/c.html) can be simplified by directly using the tooling that produces a wasi 0.2 component directly using:

```
/opt/wasi-sdk/bin/wasm32-wasip2-clang
```

vados-cosmonic:
This commit updates the C code to use the included `wasm32-wasip2-clang` binary directly for building P2 components.

Resolves #276

### DIFF
--- a/component-model/src/language-support/c.md
+++ b/component-model/src/language-support/c.md
@@ -53,9 +53,30 @@ uint32_t exports_docs_adder_add_add(uint32_t x, uint32_t y)
 }
 ```
 
-## 4. Compile a WebAssembly module (P1) with `clang`
+## 4. Compile a WebAssembly P2 component with `wasi-sdk`'s `wasm32-wasip2-clang`
 
-Now, you can compile the function into a Wasm module via clang:
+Given that `wasi-sdk` has ben installed, we can build a Preview 2 component quickly by using the provided `wasm32-wasip2-clang` binary:
+
+```console
+/opt/wasi-sdk/bin/wasm32-wasip2-clang \
+    -o adder.wasm \
+    -mexec-model=reactor \
+    component.c \
+    adder.c \
+    adder_component_type.o
+```
+
+After this command completes, you will have a new file named `adder.wasm` available in the source folder.
+
+For use cases that require building a P1 module and/or adapting an existing P1 module into a P2 module,
+such as buliding for a platform that does not support P2, details on a more manual approach can be found below:
+
+<details>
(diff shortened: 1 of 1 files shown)


component-model/src/language-support/c.md
```diff
@@ -53,7 +53,31 @@ uint32_t exports_docs_adder_add_add(uint32_t x, uint32_t y)
...
+For details on a more manual approach , see the following instructions:
+
+<details>
+<summary>Manual P1 & P2 build using `wasi-sdk`'s `clang` and `wasm-tools`</summary>
+
+## Build a WebAssembly module (P1) with `clang`
+
+> [!WARNING]
+> These instructions are no longer necessary given that `wasi-sdk`'s `wasm32-wasip2-clang` can be used directly.
+> The instructions below are left mostly as a reference for those who want to take an alternative route.
```
itowlson: By "alternate route" are we talking "fun historical recreation like spinning your own yarn" or are there circumstances in which people might need to do it this way?  If it's the former, I'd suggest removing it; if the latter, maybe we should provide guidance even if it's something as vague as "if you don't have access to a recent version of / if you need to retain backward compatibility with / etc."

itowlson (commented):
The new content looks fine (and wow, what a much smoother workflow! Nice work by the WASI SDK folks) - just a comment on keeping the old one.

component-model/src/language-support/c.md
vados-cosmonic: So people may need to make P1 components (ex. the case where they don't have access to a P2 impl), for example building for WAMR.

I'll add a comment that explains a bit more.

vados-cosmonic: Done in https://github.com/bytecodealliance/component-docs/pull/289/commits/b0e4bf1e00a4726630a6ead6afbea96a82bfca53 ! Let me know what you think

itowlson: The main text looks good - but this "warning" in the expanded detail now feels a bit weird with the "don't do this unless you want to."  Maybe use similar phrasing here as on line 71?  Not a blocker for sure

itowlson (approved):

component-model/src/language-support/c.md
vados-cosmonic: Ah thanks this makes sense, I'll update this!

vados-cosmonic: I ended up removing the warning all together

PR #296, opened 2025-08-04
fix(lang/py): componentize py usage

vados-cosmonic:
This commit fixes a few issues with `componentize-py` in the Python docs and fixes the docs as well, removing a no-longer-used section.

### DIFF
--- a/component-model/src/language-support/python.md
+++ b/component-model/src/language-support/python.md
@@ -52,7 +52,7 @@ in a file called `app.py`:
 ```py
 from wit_world import exports
 
-class Add(exports.Adder):
+class Add(exports.Add):
     def add(self, x: int, y: int) -> int:
         return x + y
 ```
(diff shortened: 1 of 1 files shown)


component-model/src/language-support/python.md
```diff
@@ -15,34 +15,44 @@ Next, create or download the WIT world you would like to target. For this exampl
...
+
+To implement the `adder` world (in particular the `add` interface that is exported), put the following code
+in a file called `app.py`:
 
 ```py
-import adder
+from wit_world import exports
 
-class Add(adder.Adder):
+class Add(exports.Adder):
```
womeier:
```suggestion
class Add(exports.Add):
```

I've tested it [here](https://github.com/womeier/wasm-components-playground/tree/main/python), and it works otherwise. Thanks!

vados-cosmonic: Ah thanks -- I didn't copy that over properly from my local :)

itowlson (approved):
I'm not sufficiently familiar with `componentize-py` to verify on correctness, but if you're okay with that, then have at it!

PR #299, opened 2025-08-12
Revise C# language support section

catamorphism:
Added more details where necessary. Factored code examples out into separate files and included them.

Fixed a bug where the package name in the Program.cs file (in the "host-app" project) didn't match the package name in the .wit

Factored out duplicated text in "Run the component from the example host" section that also appears in the C section; included this text from separate .md files.

### DIFF
--- a/component-model/src/language-support/csharp.md
+++ b/component-model/src/language-support/csharp.md
@@ -38,11 +38,12 @@ cd adder
 
 Next, create or download the WIT world you would like to target.
 
-For this example we will use the [`adder` world][adder-world], with an `add` function.
+For this example we will use a WIT file containing two worlds
+(we'll only use the `example` world at first).
 Copy and paste the following into a new file called "`wit/component.wit`".
 
 ```wit
-{{#include ../../examples/tutorial/wit/adder/world.wit}}
+{{#include ../../examples/tutorial/csharp/adder/world-hostapp.wit}}
 ```
 
 In the `adder.csproj` project file, add a new `<ItemGroup>`
@@ -50,7 +51,7 @@ at the same level as the existing `<ItemGroup>`:
 
 ```xml
 <ItemGroup>
-    <Wit Update="wit/component.wit" World="adder" />
+    <Wit Update="wit/component.wit" World="example" />
 </ItemGroup>
 ```
 
@@ -68,8 +69,6 @@ And remove the automatically generated `Program.cs` file:
 rm Program.cs
 ```
 
-[adder-world]: https://github.com/bytecodealliance/component-docs/tree/main/component-model/examples/tutorial/wit/adder/world.wit
-
 ## 3. Write the implementation for the `adder` world
 
 If you try to build the project with `dotnet build`, you'll get an error like the following:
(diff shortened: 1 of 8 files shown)


catamorphism:
Note: I'm on vacation until September 2; will see any comments when I return.

component-model/src/language-support/csharp.md
```diff
@@ -36,29 +38,23 @@ cd adder
...
+at the same level as the existing `<ItemGroup>`:
 
 ```xml
 <ItemGroup>
     <Wit Update="wit/component.wit" World="adder" />
 </ItemGroup>
 ```
 
-Since this component will only export a function dotnet considers this a library project.
+Since this component will only export a function, dotnet considers this a library project.
```
vados-cosmonic:
```suggestion
Since this component will only export functionality, dotnet considers this a library project.
```

component-model/src/language-support/csharp.md
```diff
@@ -112,90 +100,50 @@ dotnet build
...
 
 ## Building a component that exports an interface
 
-The previous example uses a WIT file that exports a function. However, you'll often prefer to export an interface,
-either to comply with an existing specification or to capture a set of functions and types that tend to go
-together. Let's expand our `example` world to export an interface rather than directly
-export the function. We are also adding the `hostapp` world to our WIT file which we will implement
-in [the next section](#building-a-component-that-imports-an-interface) to demonstrate how to build a
-component that *imports* an interface.
+The previous example uses a WIT file that exports a function.
```
vados-cosmonic: This section is lo longer true, right? We always export an interface now

component-model/src/language-support/csharp.md
```diff
@@ -301,6 +236,11 @@ wac plug \
     -o main.wasm
 ```
 
+If you get an error message like "error: the socket component had no matching imports
```
vados-cosmonic: I think we could put this error in a code block if we know what it looks like when output to the console

vados-cosmonic (commented):
Thanks for the update here!

component-model/src/language-support/csharp.md
catamorphism: Hmm, good point. Should the previous example even be included, then?

component-model/src/language-support/csharp.md
catamorphism: Fixed in [86744f1](https://github.com/bytecodealliance/component-docs/pull/299/commits/86744f1921d47857cd5a9cbf5e188408c9adf894)

component-model/src/language-support/csharp.md
vados-cosmonic: Yeah IMO we can just remove this entire section, what do you think? I think we've done it before for other guides, where once we updated to using an interface by default we could just chop the section that showed it as a "next step".

catamorphism: SGTM -- removed it in [776c8d7](https://github.com/bytecodealliance/component-docs/pull/299/commits/776c8d7e4c50e66e7ca720096b4569b0f102f831), let me know what you think now.

vados-cosmonic: So I was a bit surprised because the world is now `example` and we can't just depend on the adder example/have that match other guides closely, but I see that's required because the example goes on to show people how to import as well.

Maybe some day we'll split up those two files into `wit/component.wit` and `wit/host.wit` or `wit/embedder.wit` or something like that. For now this looks great.

PR #302, opened 2025-08-15
Replace instant with datetime in wit-example.md

og-syn:
I would imagine what happened here is that what is currently defined as datetime used to be called instant. It looks like it got renamed in some places, but not everywhere.

This patch tries to take care of that.

Apologies if I'm messing something up, this is my first pull request ever in this project.

### DIFF
--- a/component-model/src/design/wit-example.md
+++ b/component-model/src/design/wit-example.md
@@ -79,13 +79,13 @@ an unsigned 32-bit integer.
 The following declares a function named `now`:
 
 ```wit
-now: func() -> instant;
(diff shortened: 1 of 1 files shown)


og-syn:
No problem. I'm very unfamiliar with the processes to get PRs merged in this project (I'm new to webassembly in general, hence my reading the docs) -- if there's something that I should do / know to help please feel free to refer me to documentation. I looked at the code of conduct and the contributing guidelines but they seem pretty generic for this project.

catamorphism:
You're doing everything right as far as I can tell!

PR #303, opened 2025-08-15
'add' to 'foo' and minor typo in wasmtime.md

og-syn:
This fixes an example which uses 'add' but where the surrounding text uses 'foo' as a function name. Foo seems more thematic.

There's also a minor typo ('and' is misspelled AFAICT).

### DIFF
--- a/component-model/src/running-components/wasmtime.md
+++ b/component-model/src/running-components/wasmtime.md
@@ -32,9 +32,9 @@ Try out building and running HTTP components with one of these tutorials
 As of Wasmtime Version 33.0.0, there is [support for invoking components with custom exports](https://bytecodealliance.org/articles/invoking-component-functions-in-wasmtime-cli).
 
 
-As an example, if your component exports a function `foo` which takes two numeric arguments, you can make use of this feature with the following command.
+As an example, if your component exports a function `add` which takes two numeric arguments, you can make use of this feature with the following command.
 
 ```sh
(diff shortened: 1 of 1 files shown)


component-model/src/running-components/wasmtime.md
```diff
@@ -32,9 +32,9 @@ Try out building and running HTTP components with one of these tutorials
 As of Wasmtime Version 33.0.0, there is [support for invoking components with custom exports](https://bytecodealliance.org/articles/invoking-component-functions-in-wasmtime-cli).
 
 
-As an example, if your component exports a function `add` which takes two numeric arguments, you can make use of this feature with the following command.
+As an example, if your component exports a function `foo` which takes two numeric arguments, you can make use of this feature with the following command.
```
vados-cosmonic: I think the example here should probably use `add` rather than `foo` -- I think the original intent here was to use a function that is both easy to understand at a glance and something related to the guides (where we create an adder component).

vados-cosmonic (commented):
Thanks for fixing the typo @og-syn -- left a comment about the function, I think `add` was probably the right thing, and we should instead change the `foo` invocation.

component-model/src/running-components/wasmtime.md
og-syn: Sounds good. I'll mess with github a bit and see if I can update the PR or add a patch to it. Thanks for looking into this!

PR #309, opened 2025-08-29
feat: document code fragment rules

### LINKED ISSUE #239: Consistent rules for executable code fragments
After the meeting of SIG Docs today, we landed on what are consistent rules for dealing with syntax highlighting for *commands*. 

While there has been discussion (and there are always opinions!) the principles that are in use at Fermyon seem to be good enough and prevent bikeshedding:

- `console` is used only when commands *and* output are shown
  - `$` must be before lines that show commands to show that a shell is being used
- `sh`/`bash` is for commands that should be copy-pastable

While there are many ways this *could* go, I think standardizing on this usage and encoding it in CONTRIBUTING.md (CONVENTIONS.md?)or some such file is important to move past this quickly and be consistent going forward!

vados-cosmonic:
This commit adds a basic style guide to the contributing documentation, with rules around how we currently do executable code fragments.

Resolves #239

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -6,7 +6,7 @@ The Component Model documentation is a [Bytecode Alliance](https://bytecodeallia
 
 You can run the website locally using the [mdBook](https://rust-lang.github.io/mdBook/index.html) command line tool.
 
-## Prerequisites
+### Prerequisites
 
 To use this repository, you need [mdBook](https://rust-lang.github.io/mdBook/guide/installation.html) installed on your workstation.
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -6,7 +6,7 @@ The Component Model documentation is a [Bytecode Alliance](https://bytecodeallia
 
 You can run the website locally using the [mdBook](https://rust-lang.github.io/mdBook/index.html) command line tool.
 
-### Prerequisites
+## Prerequisites
 
 To use this repository, you need [mdBook](https://rust-lang.github.io/mdBook/guide/installation.html) installed on your workstation.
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -6,7 +6,7 @@ The Component Model documentation is a [Bytecode Alliance](https://bytecodeallia
 
 You can run the website locally using the [mdBook](https://rust-lang.github.io/mdBook/index.html) command line tool.
 
-### Prerequisites
+## Prerequisites
```
itowlson: This is part of "Using this repository" so the H3 markdown was correct.

itowlson (commented):
The style guide looks fine but there are some unrelated changes which I think are not correct.

CONTRIBUTING.md
vados-cosmonic:
```suggestion
### Prerequisites
```

Thanks!

CONTRIBUTING.md
```diff
@@ -20,7 +20,7 @@ cargo install --version 0.7.7 mdbook-linkcheck
 
 [cargo]: https://doc.rust-lang.org/cargo
 
-### Running the website locally
+## Running the website locally
```
itowlson: Sorry, I didn't comment on this line specifically before but I think this is still part of "Using this repository."  So my suggestion for this PR would be to revert this change.

That said, "Using this repository" is a pretty open-ended title and perhaps we should be considering how to re-structure this anyway...!

vados-cosmonic: Ah thank you! Will update this one too :)

```suggestion
### Running the website locally
```

PR #311, opened 2025-09-05
Revise Go language support section

catamorphism:
Added more details where necessary. Factored code examples out into separate files and included them.

Made the "Testing the add component" section consistent with other examples involving the example host.

### DIFF
--- a/component-model/src/language-support/go.md
+++ b/component-model/src/language-support/go.md
@@ -84,8 +84,9 @@ so it requires the imports of `wasi:cli/imports@0.2.0`.
 
 Tools like [`wkg`][wkg] can be convenient to build a complete WIT package by resolving the imports.
 
-Running the `wkg wit build` command will resolve the imports
-and populate your `wit` folder with all relevant imported namespaces and packages.
+Running the `wkg wit build` command encodes the WIT into the Component Model binary format.
+As a side effect, it resolves the imports
+and populates your `wit` folder with all relevant imported namespaces and packages.
 
 ```
 $ wkg wit build


component-model/src/language-support/go.md
```diff
@@ -60,41 +64,28 @@ Consider also running `go mod tidy` after adding the above tool.
...
+(part of [`wasi:cli`][wasi-cli]),
+so it requires the imports of `wasi:cli/imports@0.2.0`.
 
 ### Using `wkg` to automatically resolve and download imports
 
 Tools like [`wkg`][wkg] can be convenient to build a complete WIT package by resolving the imports.
 
-Running the `wkg wit fetch` command will resolve the imports and populate your `wit` folder with all relevant
-imported namespaces and packages.
+Running the `wkg wit build` command will resolve the imports
```
vados-cosmonic: I think this should actually be `fetch`... Am I wrong here? `fetch` is the one that resolves dependencies. They probably both work, I wonder if we should lean towards `fetch` as it makes more semantic sense.

component-model/src/language-support/go.md
```diff
@@ -254,50 +245,54 @@ internal
...
-We can build our component using TinyGo by specifying the wit-package to be `add.wit` and the WIT world to be `adder`.
-
-Under the hood, TinyGo invokes `wasm-tools` to embed the WIT file to the module and componentize it.
+We can build our component using TinyGo.
+Under the hood, TinyGo invokes `wasm-tools`
+to embed the WIT file to the module and componentize it.
 
 ```console
-tinygo build -target=wasip2 -o add.wasm --wit-package docs:adder@0.1.0.wasm --wit-world adder main.go
+tinygo build -target=wasip2 -o adder.wasm \
```
vados-cosmonic:
```suggestion
tinygo build -target=wasip2 \
    -o adder.wasm \
```

component-model/src/language-support/go.md
catamorphism: I think in the context of the tutorial, `build` makes more sense since it also fetches dependencies and it's necessary in order to do the next step?

vados-cosmonic: yeah what is weird is that it doesn't actually... build anything, right? What it really does is look at the dependencies and then fetch them. 

I think of it like an `npm install` vs `npm build` (if `npm build` was even built in) -- clearly one is preparing and the other is building, but `wkg wit build` doesn't really seem to "build"

itowlson: @vados-cosmonic I thought `wit wkg build` built the WIT into the CM binary format.  (So that it could be pushed to a registry for distribution.)  The fetching is required for the building, but is a side effect rather than the purpose.

vados-cosmonic: Ah thanks for pointing that out -- this guide is unique in that it uses the WIT-only wasm component w/ `go tool wit-bindgen-go` -- I didn't realize this is the only guide that does that. 

In that case we probably want to update the description of what the command does, because the current description matches what `wkg wit fetch` does, not what `wkg wit build` does -- it's weird that we would build something to be pushed to a registry for distribution but the commands that come later use that WIT-only component...

I personally think `wit/deps` is much more grokkable than comitting a binary BUT maybe the intended goal/ideal state is to not commit the binary at all and people to run this locally before they start working? Either way that's probably outside the scope of this discussion/PR. At the very least, the description of what this is supposed to do (and why) is not really helpful/isn't accurate if we want to do `wkg wit build`,

catamorphism: The next step definitely requires `build` to be run and not just `fetch`, so I changed the description of `build` -- take a look at [b55054c](https://github.com/bytecodealliance/component-docs/pull/311/commits/b55054c57aa6dbd9d7dc442250d6cd569820d1c9) and let me know if that looks good?

component-model/src/language-support/go.md
```diff
@@ -60,41 +64,29 @@ Consider also running `go mod tidy` after adding the above tool.
...
+(part of [`wasi:cli`][wasi-cli]),
+so it requires the imports of `wasi:cli/imports@0.2.0`.
 
 ### Using `wkg` to automatically resolve and download imports
 
 Tools like [`wkg`][wkg] can be convenient to build a complete WIT package by resolving the imports.
 
-Running the `wkg wit fetch` command will resolve the imports and populate your `wit` folder with all relevant
-imported namespaces and packages.
+Running the `wkg wit build` command encodes the WIT into the Component Model binary format.
```
itowlson, line 87: The title of this section says it's about "resolve and download imports" so as a reader I have questions about why I am doing this encoding malarkey.  After all, "resolve and download" is what `fetch` does, so why am I not doing that?

Looking ahead to section 3, I am _guessing_ it is something like: in order to use `go tool wit-bindgen-go`, we must have a package (that is addressable using package notation) rather than a WIT text tree: therefore we must do a wit build rather than a wit fetch.

Is that correct?  That `go tool wit-bindgen-go` can't operate on the text tree, only on a binary package reference?  I'm a bit surprised by that but the one thing I do know about Go is its infinite capacity to surprise me! _grin_

If that is the reason, I'd suggest explicitly explaining this context, because reading this now the build step feels a bit arbitrary, likely as a result of partial updates as the Go tooling has changed over time...

PR #314, opened 2025-09-05
Edit "Composing and Distributing Components" section for clarity

catamorphism:
This is a collection of minor edits, plus adding semantic line breaks. This probably makes it hard to see where the changes are, since they're minor -- apologies in advance -- but hopefully the semantic line breaks will make these files easier to edit going forward!

### DIFF
--- a/component-model/src/composing-and-distributing/composing.md
+++ b/component-model/src/composing-and-distributing/composing.md
@@ -6,7 +6,7 @@ The component model accomplishes this by packaging code in a portable binary for
 and providing  machine-readable interfaces in [WIT](../design/wit.md)
 with a standardised Application Binary Interface (ABI).
 In the same way that, for example, a Rust package (crate) can be compiled together with other Rust code
-to create a higher-level library or an application, a WebAssembly component can be linked with other components.
+to create a higher-level library or an application, a WebAssembly component can be composed with other components.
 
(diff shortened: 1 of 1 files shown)


component-model/src/composing-and-distributing/composing.md
```diff
@@ -1,115 +1,157 @@
...
-
-> Component model interoperation is more convenient and expressive than language-specific foreign function interfaces. A typical C FFI involves language-specific types, so it is not possible to link between arbitrary languages without at least some C-language wrapping or conversion. The component model, by contrast, provides a common way of expressing interfaces, and a standard binary representation of those interfaces. So if an import and an export have the same shape, they fit together directly.
+The WebAssembly component model enables applications and components to work together,
+no matter what languages they were originally written in.
+The component model accomplishes this by packaging code in a portable binary format
+and providing  machine-readable interfaces in [WIT](../design/wit.md)
+with a standardised Application Binary Interface (ABI).
+In the same way that, for example, a Rust package (crate) can be compiled together with other Rust code
+to create a higher-level library or an application, a WebAssembly component can be linked with other components.
```
vados-cosmonic:
```suggestion
to create a higher-level library or an application, a WebAssembly component can be composed with other components.
```
I think the word "compose" might be better here, because linking exists and is a slightly different thing

PR #320, opened 2025-10-01
Emscripten does not support the component model

MendyBerger:
@tschneidereit wrote on the [BA Zulip](https://bytecodealliance.zulipchat.com/#narrow/channel/206238-general/topic/Emscripten.20Component.20Model.20Support/near/542487871) 

> Emscripten only supports wasip1


vados-cosmonic:
Yep, there are docs that mention using preview1 however, so it's nto completely out of scope, see [this section](https://component-model.bytecodealliance.org/language-support/c.html?highlight=preview1#optional-build-a-wasi-enabled-webassembly-p2-component-with-wasm-tools). 

We in-general have to be careful with just pretending that previous ways of doing things never existed, so striking the balance between pushing people to more aligned approaches and leaving something recognizable was the goal here.

PR #350, opened 2026-06-12
Add WASI P3 tabs to creating-runnable-components/rust.md

ericgregory:
New P3 tab covers the library/reactor pattern with `async fn run()`


vados-cosmonic (approved):
So I think the version pinning section here is no longer true but let's get that updated in a follow up

PR #351, opened 2026-06-12
Add Async, Streams, and Futures concepts page

ericgregory:
Adds a new `design/async.md` introducing the Canonical ABI primitives (`async func`, `stream<T>`, `future<T>`) that the Component Model added for WASI P3, covering the sandwich problem, each primitive with a WIT example, the stream-plus-future and write-direction-flip patterns, and tooling pointers; hooks it into Component Model Concepts as a new subsection after Packages.

### DIFF
--- a/component-model/src/SUMMARY.md
+++ b/component-model/src/SUMMARY.md
@@ -11,6 +11,7 @@
   - [Interfaces](./design/interfaces.md)
   - [Worlds](./design/worlds.md)
   - [Packages](./design/packages.md)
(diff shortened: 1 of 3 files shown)


ericgregory:
Closing in favor of #352, which now bundles this page with the Migrating from WASI P2 to WASI P3 guide. The two pages cross-reference each other and form a coherent piece of P3 conceptual content; bundling avoids stack-management overhead and guarantees the cross-links land together.

PR #358, opened 2026-06-17
Clarify that all wit files must belong to some package

jingibus:
This seems to be implied; otherwise, where do the definitions in the file live?

### DIFF
--- a/component-model/src/design/wit.md
+++ b/component-model/src/design/wit.md
@@ -537,8 +537,8 @@ As with `use` directives, you can `include` worlds from other packages.
 ## Packages
 
 A package is a set of interfaces and worlds,
-potentially defined across multiple files.
-All files are contained in some package, 
+potentially defined across multiple files in the same directory.
+Each WIT file is associated with one package, 
 which is declared either directly in that file or in a peer file in the same directory.
 
 To declare a package, use the `package` directive to specify the package ID.
--- a/component-model/src/design/wit.md
+++ b/component-model/src/design/wit.md
@@ -538,6 +538,9 @@ As with `use` directives, you can `include` worlds from other packages.
 
 A package is a set of interfaces and worlds,
 potentially defined across multiple files.
+All files are contained in some package, 
+which is declared either directly in that file or in a peer file in the same directory.
+
 To declare a package, use the `package` directive to specify the package ID.
 A package ID must include a namespace and name, separated by a colon,
 and may optionally include a [semver](https://semver.org/)-compliant version number:


component-model/src/design/wit.md
```diff
@@ -538,6 +538,9 @@ As with `use` directives, you can `include` worlds from other packages.
 
 A package is a set of interfaces and worlds,
 potentially defined across multiple files.
+All files are contained in some package, 
+which is declared either directly in that file or in a peer file in the same directory.
```
vados-cosmonic:
```suggestion
potentially defined across multiple files in the same directory.
All files are contained in some package, 
which is declared either directly in that file or in a peer file in the same directory.
```
Thanks for the contribution @jingibus , what do you think about the above?

It feels like the last sentence could also be clearer... "files" don't go in a "package" but rather in a "folder" so it might be somewhat confusing to mix the metaphors there.

For the last sentence does this seem clearer? 

> Type, interface and world definitions must be part of a package,
declared in one or more files in a given directory.

Maybe that isn't an improvement actually...

@jingibus I do agree that this probably needs to be called out though -- would you mind sharing the surprising behavior you ran into? What did you try to do that didn't work/was surprising? Maybe we can figure out a clearer way to avoid the pitfall together

jingibus: > Thanks for the contribution @jingibus , what do you think about the above?

I think that helps a lot!

> It feels like the last sentence could also be clearer... "files" don't go in a "package" but rather in a "folder" so it might be somewhat confusing to mix the metaphors there.

It's a tricky thing to communicate! 

The hole that was bugging me was, "Can you have a WIT file in a directory with no package at all, in it or any other file in that folder?"

And the simplest answer to that question, in my mind, is "Well, WIT files are containers for WIT stuff, and WIT stuff lives in packages, and thus so must the files as well."

I think most folks will get that! Packages and folders are 1:1, after all, so it's at least unambiguous whichever word we use. But saying "files live in a package" makes that relationship clear.

> I do agree that this probably needs to be called out though -- would you mind sharing the surprising behavior you ran into? What did you try to do that didn't work/was surprising? Maybe we can figure out a clearer way to avoid the pitfall together

Oh, I ran into the ambiguity as I was working on validation for a WIT parser (https://github.com/wasmcomputercompany/brevity). I suspect that the existing implementations are perfectly clear — after all, it's not like you can have a world or an interface defined without a package.

jingibus: I've included the suggested edit, and tweaked the verbiage to try and address those concerns about mixing metaphors.

vados-cosmonic: A bit late here but thanks for giving his some more thought!

> The hole that was bugging me was, "Can you have a WIT file in a directory with no package at all, in it or any other file in that folder?"

Yeah, this is an important gotcha that is good to spell out. The idea that packages are ... roughly folders is something that. While I don't think the relationship is in stone it is, for all intents and purposes now and will probably be difficult to change.

> I think most folks will get that! Packages and folders are 1:1, after all, so it's at least unambiguous whichever word we use. But saying "files live in a package" makes that relationship clear.

Yeah I mostly agree -- the programming idea of a 'package' is annoyingly closely situated here. 

> Oh, I ran into the ambiguity as I was working on validation for a WIT parser (https://github.com/wasmcomputercompany/brevity). I suspect that the existing implementations are perfectly clear — after all, it's not like you can have a world or an interface defined without a package.

Thanks for noting this (and cool project! -- really excited with all the kotlin movement lately!) -- yeah, the restrictions on how to lay out a package are often the most frustrating, and how the `deps` dir is a special case, essentialy...

vados-cosmonic: Going to get this in, but maybe someday we can even have some little dir trees (like `tree` output) that spell out what you can/can't do

vados-cosmonic (approved):
LGTM 🚀 

Thanks again for contributing this @jingibus !

component-model/src/design/wit.md
jingibus: > yeah, the restrictions on how to lay out a package are often the most frustrating, and how the deps dir is a special case, essentialy...

Oh TIL. I'll have to stay on lookout for that.

vados-cosmonic: Yeah so basically `wit/deps` is almost always there as a folder, 

I suspect you've already taken a look at `wkg` -- but just in case you haven't: 
https://github.com/bytecodealliance/wasm-pkg-tools

Lots of the idiosyncracies are spelled out there (the code is a bit messy at times due to various changes in how packaging works, etc), and of course there is wit parser's `Resolve` which is where a lot of the magic happens (in particular, [`parse_deps_dir`](https://github.com/bytecodealliance/wasm-tools/blob/e6317b4f33aa5867ee902623f5e57cfea0733035/crates/wit-parser/src/resolve/fs.rs#L138))

PR #363, opened 2026-07-29
docs(language-support/rust): update wit-bindgen syntax and resource t…

gm305:
- Added "generate_all" to "wit_bindgen::generate!()" to properly generate code for imported interfaces
- Calling "bindings::export!(Component)" externally now fails with "cannot find export in self", so i deleted it
- Updated trait implementation paths to include the "exports" submodule


vados-cosmonic (approved):
LGTM 🚀 

Thanks for the improvements here @gm305 !

PR #373, opened 2026-09-09
fix: wasi preview2 links

### LINKED ISSUE #372: Broken links @https://component-model.bytecodealliance.org/
I was reading here https://component-model.bytecodealliance.org/   and you have a link here https://github.com/WebAssembly/WASI/blob/main/docs/Preview2.md wich does not work. In that document it said if i find a problem let you know .

vados-cosmonic:
Resolves #372 , enables external link checking for every commit from now on... 

I think we had it disabled to avoid pinging external servers from CI/locally when running tests, but catching dead external links (especially when it's to bytecodealliance sites) seems too important to skip. We can likely revisit if it becomes a problem


component-model/src/language-support/importing-and-reusing-components/javascript.md
```diff
@@ -35,7 +35,7 @@ which contains the `reverse-and-uppercase` function (in JavaScript, `reverseAndU
 > Functionality is imported via the `interface`, *not* the `world`.
 > `world`s can be included/used, but the syntax is slightly different for that.
 
-The JavaScript to make this work ([`string-reverse-upper.mjs` in `jco/examples`][string-reverse-upper-mjs])
+The JavaScript to make this work ([`string-reverse-upper.js` in `jco/examples`][string-reverse-upper-js])
```
mkatychev, line 38: Did these get converted at some point or were always `.js`?

mkatychev (approved):
LGTm

.github/actions/mdbook/action.yml
```diff
@@ -70,6 +70,7 @@ runs:
       id: book-build
       env:
         PUBLISH_DOMAIN: ${{ inputs.publish-domain }}
+        MDBOOK_OUTPUT__LINKCHECK2__FOLLOW_WEB_LINKS: 'true'
```
mkatychev, line 73: If we wanted to reduce the amount of times we are hitting upstream links we could make enable merge queue and only trigger [`on: merge_group:`](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-a-merge-queue#triggering-merge-group-checks-with-github-actions) by checking the [event trigger](https://docs.github.com/en/actions/reference/workflows-and-actions/contexts#github-context:~:text=for%20GitHub%20Actions%2E-,github%2Eevent,-object).

```yaml
env:
  MDBOOK_OUTPUT__LINKCHECK2__FOLLOW_WEB_LINKS: ${{ github.event == 'merge_group' && 'true' || 'false' }}
```

component-model/src/language-support/importing-and-reusing-components/javascript.md
vados-cosmonic: evidently [not](https://github.com/bytecodealliance/jco/blob/main/examples/components/string-reverse-upper/string-reverse-upper.js)

.github/actions/mdbook/action.yml
vados-cosmonic: As discussed in the meeting this will be handled in a follow up

